PR #39, opened 2015-05-22
Changed UpcomingEvents logic

AussieInSeattle:
With logic change an Event in "Upcoming Events" will stay on the list until they have finished (now using end date).  Additionally, since the server datetime seems to be set to somewhere in Europe (+9hrs of Seattle) we show events on the list up to 1 day after they have finished to get around any timezone differences.

Note: please confirm the property alias of the end date is actually "end" - did not see anywhere in the repo where I could see that?


nul800sebastiaan:
Thanks and sorry for the late reply. We'll bring back the events soon and look into this, but that will be new code. So I'm closing this for now!

PR #44, opened 2015-06-15
Our 129

bootnumlock:
not sure why the project forums are hidden-xs as they would be the MOST important feature for a mobile user as they cannot download or install the package from the smallest devices, right?


nul800sebastiaan:
Well, you sent a pull request :+1: but it didn't contain any of your commits :bowtie:. I think you wanted to create a pull request for these differences? https://github.com/umbraco/OurUmbraco/compare/v7-upgrade-test...bootnumlock:v7-upgrade-test
Please try again! :sparkles: :boom:

PR #45, opened 2015-06-15
Our 129

bootnumlock:
not sure why the project forums are hidden-xs as they would be the MOST important feature for a mobile user as they cannot download or install the package from the smallest devices, right?


nul800sebastiaan:
Thanks! :clap::clap::clap:

greystate:
:zap: :rocket:

PR #48, opened 2015-06-19
move ms-tiles into xml + fix paths for tile images

ceee:


nul800sebastiaan:
I trust you, I have no idea how this works.. ;-)

ceee:
:) It's just a hint for IE where to find the configuration file, see [here](https://msdn.microsoft.com/en-us/library/dn320426%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396).

I have moved it into the xml so the HTML doesn't get polluted with the metadata. Only IE will download the file.

PR #53, opened 2015-06-27
Fix for OUR-257 (Report Compatibility display bug in Safari)

greystate:
Fix for a display bug in Safari.


nul800sebastiaan:
Awesome, thanks!

PR #55, opened 2015-07-01
Typos OUR 272 and OUR 273

greystate:
Just a couple of quick fixes for the mentioned issues.


greystate:
Should I submit these as separate PRs instead of combining them? Seemed too tiny to warrant their own :)

PR #57, opened 2015-07-05
Fix: OUR-279: Signup page show a white header when opening quickmenu

Aaen:
Change css so  the header is not missing when you are on the signup page, when opening the quickmenu

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,3 +59,4 @@ npm-debug.log
 OurUmbraco.Site/[Ww]eb.config
 OurUmbraco.Site/[Aa]ssets/css/style.min.css
 OurUmbraco.Site/[Aa]ssets/js/app.min.js
+.vs/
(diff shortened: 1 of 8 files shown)


nul800sebastiaan:
Hey Dennis! Can you please try this again, not sure what went wrong here but that looks like way too many commits.

Best to start over completely, so: 
* Delete your fork at https://github.com/Aaen/OurUmbraco/settings
* Fork from this repository again
* The very easiest way to do this would be to go to https://github.com/Aaen/OurUmbraco/blob/master/OurUmbraco.Client/src/scss/elements/_quickmenu.scss and hit the "Edit" button
* Make your edits
* In the bottom you'll see the commit dialog, including a radiobutton that says: `Create a new branch for this commit and start a pull request.` 
* Follow the instructions and :tada:

PR #62, opened 2015-07-28
OUR-304: Added font-family to release issues list

jannikanker:
http://issues.umbraco.org/issue/OUR-304


jannikanker:
Bollocks, that removes the icons! Never ever mind :-(

PR #63, opened 2015-07-29
Fix: OUR-304

jannikanker:
Added font-family property to .markdown-syntax anchor tags


nul800sebastiaan:
Much better, thanks! This is now rolled out to live!

PR #65, opened 2015-08-04
OUR-310: Fix Compact view on home if logged in

Aaen:
If a user is logged in on Our, we should have a more compact view of the home page - could easily be done by 
 removing the massive whitespace around search
 remove the about text 
 move forum posts above the two documentation promos


nul800sebastiaan:
Great job!

PR #68, opened 2015-08-17
OUR-322: Removed the is Stable information, you can´t add it on the c…

Aaen:
If you go to a specific package page (for example: https://our.umbraco.org/projects/backoffice-extensions/perplexmail-for-umbraco/) you see on the rightside "Is stable: False". There used to be a checkbox for that option in the edit-screen of a package but that disappeared.

I think the line that displays this should be removed completely (, or there has to be a checkbox on the edit-page)


nul800sebastiaan:
:+1:

PR #71, opened 2015-08-21
Correctly pluralize "votes", using a helper

greystate:
This is a quick fix for OUR-318, using a `pluralize()` helper. Not sure if there's a "general" place to put said helper - let me know if it's preferred to be in a specific file.


nul800sebastiaan:
:dancers:

PR #72, opened 2015-08-21
Change from HTTP to HTTPS in notification links

greystate:
Fix for OUR-327 - only by replacing the actual strings in the code.


nul800sebastiaan:
Great! Not sure when this will be deployed, but should be soon!

greystate:
No worries - awesome!

PR #80, opened 2015-10-01
Emphasizing the current Umbraco version (when linked to)

greystate:
Easter-egg for @nul800sebastiaan to link to when announcing that #Our is running the latest version :-)

_Reference screenshot:_

umbraco-version


greystate:
Just noticed I didn't mention this other than in code, but the point is, of course, that the highlight in the screenshot only shows when you're specifically linking to the _umbraco-version_ ID on the page :)

PR #82, opened 2015-11-19
Ensure GitHub URLs are always in correct case.

ed-parry:
This is to resolve [Issue #271](https://github.com/umbraco/UmbracoDocs/issues/271) within the UmbracoDocs repo. Issue being users navigating the documentation site with lowercase URLs were throwing 404's when hitting GitHub which is case sensitive.

First contribution to Umbraco, so apologies in advanced if my processes aren't correct!

### DIFF
--- a/OurUmbraco/Documentation/DocumentationContentFinder.cs
+++ b/OurUmbraco/Documentation/DocumentationContentFinder.cs
@@ -5,6 +5,7 @@ using System.Web;
 using OurUmbraco.Documentation.Busineslogic;
 using Umbraco.Core;
 using Umbraco.Web.Routing;
+using System.Text.RegularExpressions;
 
 namespace OurUmbraco.Documentation
 {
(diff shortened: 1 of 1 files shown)


hemraker:
Hi @ed-parry 

Thanks for doing this! I have tested it and it works... to an extent.. The issue is still there if the folder name contains a dash. Eg.:

`https://our.umbraco.org/documentation/Implementation/Default-Routing/`

We need to figure out some way to generate the github URL from the actual MD filesystem so we can ensure proper casing and not just make qualified guesses.

HarveyWilliams:
Can this be merged? I came across this issue today and this appears to have been outstanding for a while.

@hemraker I can't see any issues with URLs ending with a trailing slash.

HarveyWilliams:
What if we get the name of the file from the file system? Surely folders and files are named correctly on the file system, so if the file is found, rather than continuing to get the URL from the requested page, we get get the path from the file!

Shazwazza:
@HarveyWilliams yes i think that would be the right approach since the file name on disk is how it's stored in GIT which is how GitHub will respond to the request.

dampee:
@Shazwazza that is basically my fix in #97

Shazwazza:
Going to close this since we pulled in this PR https://github.com/umbraco/OurUmbraco/pull/97 and everything seems to work good now. Thanks for the effort everyone!

PR #84, opened 2016-05-12
FIX OUR-396

jannikanker:
http://issues.umbraco.org/issue/OUR-396


hartvig:
Will merge manually - h5yr!

PR #85, opened 2016-06-12
Extracts package's required Umbraco version metadata

leekelleher:
When a package is uploaded to the Our Umbraco repository, we extract the "package.xml" manifest to detect the required (minimum) Umbraco version number.

This task is part of [U4-8156](http://issues.umbraco.org/issue/U4-8156), at CGRT16


Shazwazza:
This all looks good to me

PR #41, opened 2015-06-13
OUR-148 Project releases - displayed owner's name, instead of uploader.

leekelleher:
Corrected the package uploader's name.

PR #87, opened 2016-06-13
OUR-148 Project releases - displayed owner's name, instead of uploader

leekelleher:
It turns out that I failed to fix this in #41 - I had only fixed it for the "Source code" tab, (doh!)
This now fixes it for each of the downloads tabs.


leekelleher:
Rebased my PR on latest 'master'.

@nul800sebastiaan Shall I create a new YouTrack ticket for this?

An example project where this happens would be uComponents - Shannon is created with all the uploads, when it was me. <https://our.umbraco.org/projects/backoffice-extensions/ucomponents/>

PR #97, opened 2017-01-25
OUR-437 fix casing of md file in case incoming URL has incorrect casing

dampee:
see http://issues.umbraco.org/issue/OUR-437

What this pull request changes is the way the github url is build.
With this pull request we are looking up the file on disk, then rebuild the filename (including correct casing) from the complete filepath.  This appears the only reason to get this working.
NTFS is case sensitive but the windows API's are not.  Therefor there does not seem to be another way to retrieve a properly cased file path.


dampee:
Just noticed that there is already a pull request for this: https://github.com/umbraco/OurUmbraco/pull/82

Shazwazza:
The code looks good, i'm gonna merge in and just seen what happens... YOLO!

Shazwazza:
Great work! that seems to work perfectly :)

PR #99, opened 2017-02-08
Fixed typo on Roadmap.cshtml preventing ReleaseDescription from being displayed

jamiepollock:
Very small fix which will allow the release description to be displayed instead of `release.releaseDescription`.


nul800sebastiaan:
D'oh! Thanks!! All fixed now.

PR #100, opened 2017-03-04
Our 441

stvnhrlnd:
See http://issues.umbraco.org/issue/OUR-441

### DIFF
--- a/OurUmbraco.Site/Views/Partials/Forum/Forum.cshtml
+++ b/OurUmbraco.Site/Views/Partials/Forum/Forum.cshtml
@@ -182,7 +182,7 @@
                 <a class="@Umbraco.If(i == page, "active")" href="?@queryString">@i</a>
             }
 
-            @if (page < (pages - 3))
+            @if (page < pages)
             {
                 queryString["page"] = (page + 1).ToString();
                 <span>&hellip;</span>
--- a/OurUmbraco.Site/Views/Partials/Forum/Forum.cshtml
+++ b/OurUmbraco.Site/Views/Partials/Forum/Forum.cshtml
@@ -43,9 +43,9 @@
             TimeSpan.FromSeconds(4));
     }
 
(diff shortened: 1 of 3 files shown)


hartvig:
Tested locally and works fine :-)

OurUmbraco.Site/Views/Partials/Forum/Forum.cshtml
```diff
@@ -168,19 +171,22 @@
...
+                <a class="@Umbraco.If(i == page, "active")" href="?@queryString">@i</a>
             }
 
-            @if (page < (pages - 3))
+            @if (page < pages)
             {
+                queryString["page"] = (page + 1).ToString();
                 <span>&hellip;</span>
-                <a class="next" href="?page=@(page + 1)">Next</a>
+                <a class="next" href="?@queryString">Next</a>
```
Shazwazza: Usages of query string directly rendering into HTML is the exact way to cause XSS issues, this will need to be fixed. I'll have a look

stvnhrlnd: Hi Shannon,

Thanks for the feedback. Did you have a look at this?

Forgive me if I'm being ignorant but won't ASP.NET protect against this? First of all any script tags in the query string will cause an `HttpRequestValidationException` ("A potentially dangerous Request.QueryString value was detected from the client"). Second, Razor will automatically encode strings output with `@` (apart from `HtmlString` of course but that's not what we're dealing with here).

I've attempted to execute an XSS attack but without any luck. Is there something I'm missing?

Cheers,

Steven

Shazwazza: Hi Steven,

This is all fine, when i glanced over the code I just saw potential XSS issues but there isn't any here and as you mention razor would encode it anyways. 
Cheers,
Shan

PR #105, opened 2017-05-05
Fixes issue #OUR-449.

naepalm:
Fixed a couple things:

- Doubled the image size on the package listing and set it smaller with the CSS so it displays nicely for retina
- Removed the border-radius on profile pages for package listings so the icons are consistent

### DIFF
--- a/OurUmbraco.Site/Views/Partials/Projects/ListProjects.cshtml
+++ b/OurUmbraco.Site/Views/Partials/Projects/ListProjects.cshtml
@@ -185,7 +185,7 @@
                                 defaultScreenshot = childContent.GetPropertyValue("defaultScreenshotPath", false, "/css/img/package2.png");
                             }
                         }
-                        <img src="@Utils.GetScreenshotPath(defaultScreenshot)?bgcolor=fff&width=50&height=50&format=png" />
+                        <img src="@Utils.GetScreenshotPath(defaultScreenshot)?bgcolor=fff&width=100&height=100&format=png" />
                     </div>
                     <div class="col-xs-10 col-md-6">
                         <div class="forum-thread-text">
--- a/OurUmbraco.Client/src/scss/elements/_packages.scss
+++ b/OurUmbraco.Client/src/scss/elements/_packages.scss
@@ -90,6 +90,7 @@ body.packages {
 			height: auto;
 			display: block;
 			vertical-align: middle;
+            width: 50px;
 			
 			@media (min-width: $md) {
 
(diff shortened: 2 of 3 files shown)


OurUmbraco.Site/Views/Partials/Projects/ListProjects.cshtml
```diff
@@ -185,7 +185,7 @@
                                 defaultScreenshot = childContent.GetPropertyValue("defaultScreenshotPath", false, "/css/img/package2.png");
                             }
                         }
-                        <img src="@Utils.GetScreenshotPath(defaultScreenshot)?bgcolor=fff&width=50&height=50&format=png" />
+                        <img src="@Utils.GetScreenshotPath(defaultScreenshot)?bgcolor=fff&width=100&height=100&format=png" />
                     </div>
```
bjarnef: I will recommend to use `srcset` attribute for this, since most devices with retina screens on Android and iOS support `srcset`. It might be an issue on desktop for people using IE.

```
<img src="@Utils.GetScreenshotPath(defaultScreenshot)?bgcolor=fff&width=50&height=50&format=png"
     srcset="@Utils.GetScreenshotPath(defaultScreenshot)?bgcolor=fff&width=100&height=100&format=png 2x,
             @Utils.GetScreenshotPath(defaultScreenshot)?bgcolor=fff&width=150&height=150&format=png 3x" />
```

naepalm: Hm, I have no problem implementing srcset, but for us Surface users, it doesn't look like Edge supports it either. I'm not sure what the stats are for Our on what browser, though. My whole team uses Surfaces (I'm on a Surface Book), but I think we all use Chrome.

It would be good to have access to that data to see who it might effect!

bjarnef: Not sure how the stats for Our in either, but I guess it is most developers. So I am not sure how many of the visitors, who actually use IE and Edge.
As always IE and Edge lack it support on newer technology compared to other browsers: http://caniuse.com/#search=srcset

However it does seem that Edge is supporting srcset in the latest versions: https://blogs.windows.com/msedgedev/2015/06/08/introducing-srcset-responsive-images-in-microsoft-edge/#XXSF2GfTD6HgteM7.97

hartvig: Edge is 2,5% - I think it'll be alright 😀

PR #112, opened 2017-06-06
Write contributors to json file and use file cached disk value

bjarnef:
Write contributors to json file and use file cached value with help from @abjerner 👍

### DIFF
--- a/OurUmbraco.Site/Views/Partials/Home/GitHubContributors.cshtml
+++ b/OurUmbraco.Site/Views/Partials/Home/GitHubContributors.cshtml
@@ -4,12 +4,10 @@
 {
     foreach (var contributor in Model.Contributors)
     {
-        var author = contributor.Author;
-        if (author != null)
-        {
-            <a href="@author.HtmlUrl" class="contributor" target="_blank" title="@author.Login">
+            <a href="@contributor.AuthorUrl" class="contributor" target="_blank" title="@contributor.AuthorLogin">
                 <div class="avatar">
-                    <img alt="@author.Login" src="@author.AvatarUrl&s=112" />
+                    <img alt="@contributor.AuthorLogin" src="@contributor.AuthorAvatarUrl&s=112"
+                         srcset="@contributor.AuthorAvatarUrl&s=224 2x, @contributor.AuthorAvatarUrl&s=336 3x" />
                     <span class="contrib-count" title="@(contributor.TotalCommits + " " + (contributor.TotalCommits == 1 ? "contribution" : "contributions"))">
                         @contributor.TotalCommits
                     </span>
@@ -22,7 +20,6 @@
                     </div>
                 </div>
             </a>
-        }
     }
     <div class="loadmore">
         <a class="button transparent" href="#" onclick="loadAllGitHubContributors(); return false;">Load more</a>
--- a/OurUmbraco/Community/Controllers/GitHubContributorController.cs
+++ b/OurUmbraco/Community/Controllers/GitHubContributorController.cs
@@ -9,6 +9,9 @@ using RestSharp;
 using Umbraco.Core.Cache;
 using Umbraco.Core.Logging;
 using Umbraco.Web.Mvc;
+using Newtonsoft.Json.Linq;
+using Newtonsoft.Json;
+using System.Text;
 
 namespace OurUmbraco.Community.Controllers
 {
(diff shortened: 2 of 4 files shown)


bjarnef:
@nul800sebastiaan can you review this one, when you have some time 👍

abjerner:
It would be awesome to have this merged, and see how it works out when there are actual visitors forcing our cache to be regularly updated. 

I do however suspect this pull request may not be enough. While it should make the statistics appear correct, the new error handling will abort fetching further statistics if at least one requests fail - which actually may be very likely. However the GitHub API documentation mentions the following:

> Computing repository statistics is an expensive operation, so we try to return cached data whenever possible. If the data hasn't been cached when you query a repository's statistics, you'll receive a `202` response; a background job is also fired to start compiling these statistics. Give the job a few moments to complete, and then submit the request again. If the job has completed, that request will receive a `200` response with the statistics in the response body.<br /><br />Repository statistics are cached by the SHA of the repository's default branch, which is usually master; pushing to the default branch resets the statistics cache.

So if GitHub doesn't have a cache for the statistics at the time we hit the API, it will respond with a `202` status code, which our code will see as an error, as we're just checking for a `200` status code. Instead - if we encounter a `202` status code - we should probably make the thread sleep for a bit to let GitHub keep up, and then request the statistics again.

I would imagine the steps for that would be something like:

1. Attempt to get statistics for each repository
    - If we receive a `202` status code for a given repository, we add that repository to a list/queue for later
2. If at least one request results in a `202` status code, we make the thread sleep for a bit (or a "few moments" as GitHub writes it)
3. Attempt to get statistics for the remaining repositories
    - Take another round if we're still receiving `202` status codes?

There are still a few uncertainties - eg.:
- how long is a "few moments"?
- can new commits/changes to the master branch force the cache to be reset while we're waiting for it to be rebuild?
- probably something I haven't thought about

Also, since we likely have to make the thread sleep, we should probably convert the code into a backtround task rather than just some code that is called asynchronously. @nul800sebastiaan we can use Hangfire for this, right?

I will try to set up some testing for this. Meanwhile, it would still be helpful to have this pull request merged.

BTW - the GitHub API documentation for reference: https://developer.github.com/v3/repos/statistics/

nul800sebastiaan:
@bjarnef I know mate, unfortunately it's very busy after CG. We'll get to this in due time. :-) Sorry for the delay!

nul800sebastiaan:
Oops not Bjarna, I mean @abjerner  :-)

nul800sebastiaan:
> Also, since we likely have to make the thread sleep, we should probably convert the code into a backtround task rather than just some code that is called asynchronously. 

We can use something like Polly to do retries, shouldn't be too much of a problem.

emmagarland:
@abjerner thanks so much - that sounds like a really good plan and agreed re the 202 and other options. 

cc @nul800sebastiaan just a quick headsup (even before this PR is merged) - GitHub contributors are currently caching within the last hour with a very limited result set, meaning that the cache needs refreshing in order to show them all again.

Looking forward to seeing the results of the PR and any additional work on this!

hartvig:
I'll review and merge this tomorrow with my changes to filter out contributions more than a year old (else the list will be super static).

emmagarland:
Hey guys sorry to bump this again - thanks for getting this out! Just noticed something and didn't know whether to raise a ticket or just write it here.

I can't see my avatar on there and it's been 24 hours and not refreshed so I think the Our Umbraco repo isn't included in the last cached results. We can force this to refresh, and check the .json file is ok.

I notice that the throwing Error line has been removed from here and @abjerner @bjarnef  I think that was included to ensure that the last cached results are used if any error happens during the GitHub sync, so removing it would mean potentially the wrong results are cached if I have understood correctly:

https://github.com/umbraco/OurUmbraco/commit/1c8fcef7541825f3cb77fc63dba22e527cf034f7

Hope this makes sense - just wanted to flag it.

Thanks!

bjarnef:
It seems it has some issues, when requesting Github API, where it might timeout in one of the requests and instead of using the existing cached data, it updates the cached results, but it missing data from one or more repositories.

image

emmagarland:
@bjarnef yes - it is currently invalid, the Github code needs forcing to run again as the list is only a few people right now.

I am pretty sure that we need to put the line where an exception is thrown _back in_ to **GitHubContributorController** _GetContributors_ method (line 103) to stop this happening, because instead of allowing invalid result set to be written to file, it will force it to read the previously "good" cache from file:

Add it here again line 103 in the controller:

` LogHelper.Warn<IGitHubContributorsModel>(string.Format("Invalid HTTP response for repository {0}", repo));
throw new Exception(string.Format("Invalid HTTP response for repository {0}", repo));`

Then GitHubGetContributorsResult method will catch any errors while pulling the data from GitHub API (line 97), and run the line:

```
  // Load the contributors from the cache (if we have any)
model.Contributors = GetCachedContributors();
```
cc @abjerner am I right? I think this was the logic we discussed at CG17?

NB: There are further optimisations to do to allow a longer period to time to pull the data back. The operation on Github API is currently pretty heavy, as it essentially gets every single week of commit history and all contributors, from all the repos combined. Also we probably want to use something like Octokit in future, that is written by Github themselves, and implement some of the suggestions from @abjerner above.

If this is right we can do a pull request to get this line back in and hopefully stop invalid data  being written to the disk cache. Then work on some further optimisations to prevent it timing out in the first place.

abjerner:
@emmagarland Yes, the error checking and the exception is essential, as we otherwise wont have the complete list if one or more requests respond with a `202` status code (or fail entirely). The line where we threw the exception was there to prevent that scenario from happening - making sure we didn't save an incomplete set of statistics. 

With my testing so far, it seems that if the cache isn't available for a given repository, we can make the thread sleep for roughly a second, and then try again, and the cache will have been populated meanwhile. So far, if a request fails, it has only been necessary with one additional retry. It is however difficult to test this, as the cache has already been populated for pretty much all the requests that I make to the API.

So far my test case looks like [this gist](https://gist.github.com/abjerner/f312d244f4bfe769a8b38b67ac2cb53a). Until now, I have only been reloading a Razor view manually - I'll try to set up some more automated tests to hopefully get more varying results.

The key issue here is getting the requests and responses right. I think we have the parsing covered, so whether we use what we have now, or switch to something like Octokit, should really not make that much a difference. Although JSON.net may be a bit better for the deserialization than RestSharp we're using now. At least I think so - I will try to test that a bit as well.

emmagarland:
Fab @abjerner. Really like the gist you posted. Thanks for confirming re the error handling - it needs to go back in next time we change this then. 

I was wondering also if the 304 response might be happening (unchanged). According to this https://developer.github.com/v3/guides/getting-started/ ":

"The 304 status indicates that the resource hasn't changed since the last time we asked for it and the response will contain no body. As a bonus, 304 responses don't count against your rate limit."

It could explain the occasional failures on live but I've never seen it happen locally yet - as I don't think it changes until a new week passes or a new commit comes along.

Anyway #h5yr for the work you've done - feels like we're getting close to solidifying this.

Shazwazza:
Hi @abjerner, @emmagarland and @bjarnef ! 

Niels is away on holidays and probably didn't get to read much of the above. Would it be possible for one of you to create a new PR that re-fixes this based on your info above? That would be super great and then i can get it merged in. Thanks!

abjerner:
@Shazwazza yes, I'm already looking into that, but doing a bit of testing with the API first. Since the Umbraco-CMS repository is a bit bigger than the other repos, it may sometimes take a bit longer for GitHub to rebuild their cache, in which case it may take three attempts to get a proper response (when using the code from the Gist I posted earlier).

So as described earlier, this should run in a background thread - I'll look into implementing this with Hangfire, and perhaps add some more sleeping time between retries - since we don't really care that much about the overall duration when it will be running in the background. That should also make sure we get a complete set of contributions :D

@emmagarland The `304` responses are only if we send an ETag header to the API with the fingerprint of the previous request, so that shouldn't be a problem for our use case.

<br /><br />

Anyways, a few of the test runs (where the cache wasn't already populated), looks like the snippets below. So far Umbraco-CMS is the only repository I've seen require a third attempt.
```
2017-07-03 11:30:01 -> === Round 1 ===

2017-07-03 11:30:02 -> Umbraco-CMS -> Accepted -> 894,7653
2017-07-03 11:30:03 -> UmbracoDocs -> OK -> 1184,0659
2017-07-03 11:30:04 -> OurUmbraco -> OK -> 814,542
2017-07-03 11:30:04 -> Umbraco.Deploy.Contrib -> OK -> 149,179
2017-07-03 11:30:04 -> Umbraco.Courier.Contrib -> OK -> 222,6117
2017-07-03 11:30:05 -> Umbraco.Deploy.ValueConnectors -> OK -> 666,7413


2017-07-03 11:30:06 -> === Round 2 ===

2017-07-03 11:30:08 -> Umbraco-CMS -> OK -> 1840,9772

2017-07-03 11:30:08 -> YAY
```

```
2017-07-04 04:45:01 -> === Round 1 ===

2017-07-04 04:45:02 -> Umbraco-CMS -> Accepted -> 786,0892
2017-07-04 04:45:02 -> UmbracoDocs -> OK -> 720,9176
2017-07-04 04:45:03 -> OurUmbraco -> OK -> 257,7039
2017-07-04 04:45:03 -> Umbraco.Deploy.Contrib -> OK -> 142,382
2017-07-04 04:45:03 -> Umbraco.Courier.Contrib -> OK -> 157,8013
2017-07-04 04:45:03 -> Umbraco.Deploy.ValueConnectors -> OK -> 206,5504


2017-07-04 04:45:04 -> === Round 2 ===

2017-07-04 04:45:04 -> Umbraco-CMS -> Accepted -> 164,1246


2017-07-04 04:45:05 -> === Round 3 ===

2017-07-04 04:45:06 -> Umbraco-CMS -> OK -> 1215,7326

2017-07-04 04:45:06 -> YAY
```

emmagarland:
Sounds fab @abjerner :) Definitely agree re a background thread

That makes sense for UmbracoCms to be the weightiest repo, and the 304 responses. 

Good shout.

Emma

abjerner:
As this pull request already has been merged, I've created another one here for the background job with improved error handling: https://github.com/umbraco/OurUmbraco/pull/120

It's still work in progress, but feel free to come with a comments ;)

emmagarland:
@abjerner fantastic! I've initially taken a look at this and I especially like the idea of the listing combining additions and deletions alongside commits, something we actually talked about last night at umBristol too as a next step. Will have a closer look on the wkend. Stellar stuff!

PR #117, opened 2017-06-23
Removes RTs from twitter feed

hartvig:


hartvig:
Issue: http://issues.umbraco.org/issue/OUR-456

PR #119, opened 2017-07-03
Fix issues with filtering projects/packages

bjarnef:
Issue: http://issues.umbraco.org/issue/OUR-460

After upgradering jQuery version from 1.7.2 to 1.9.1, because Fancybox v3 requires jQuery 1.9+, it seems to break the project filter functionality due the way to select elements `jQuery(htmlString)` versus `jQuery(selectorString)`
https://stackoverflow.com/a/23960371
https://github.com/nathanvda/cocoon/issues/127

It seems to happen, where there is no querystring, where the `getUrlVars` function on this page was returning undefined. I have modified it to return an empty array instead.

Currently the filter function works if querystring contains any values, e.g. `?version=7` or just `?test=1` (which doesn't return the console error).

### DIFF
--- a/OurUmbraco.Client/src/scss/elements/_sidebar.scss
+++ b/OurUmbraco.Client/src/scss/elements/_sidebar.scss
@@ -143,7 +143,8 @@ html {
             margin-left: 5px;
             padding-left: 10px;
             transition: max-height 300ms ease-in-out;
-            li.active > a *{
+            li.active > a *,
+            li.active > h3 {
                 color: darken($color-our, 20%);
             }
             li.open > ul {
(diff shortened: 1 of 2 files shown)


bjarnef:
For some future enhancement it could also "select" filters when having multiple parameters in querystring: https://our.umbraco.org/projects/?orderBy=popularity&version=7

But in that case I think the filter should be redesignet/improved since you can select multiple filters via the navigation (and not sure about newbies are aware of filtering via querystring). The category "Backoffice extensions" is also quite large. Maybe it could be divided into smaller groups e.g. one for "property editors" and maybe allow a project to belong to more that one category.

Previously the projects/packages had the tags property, where you could filter projects with same tag, but I think that has been removed.

For now this PR fix the javascript error.

PR #120, opened 2017-07-06
Updated the implementation for GitHub contributors

abjerner:
This is still work in progress, so it's just to put it up here for any comments.

I haven't yet had the time to test the new logic that thoroughly when integrated with Our Umbraco, but the `GitHubService` class seems to work like it should. I will do some more testing, but you can now get the overall list of contributors like this (the parameter is the amount of retries, where `3` seems to be a fitting number):

```C#
GitHubService service = new GitHubService();
GitHubContributorsResult result = service.GetOverallContributors(3);
```

The `GetOverallContributors` also sorts the contributors a bit better - eg. if two users have the same amount of commits, but one has more additions than the other. So the one with the highest amount of additions is shown first. I'm also sorting by deletions should two users have the exact same amount of commits and additions - even though that is probably an edge case.

The actual view showing the contributors is calling the `GitHubContributorController` surface controller. I haven't updated this with the new logic yet, but that's probably the next thing I'm going to work on.

Since we want to update the contributors in the background, I've configured a Hangfire recurring job to call the `UpdateOverallContributors` method once a day.  The logic for updating the JSON is isolated in it's own method in the service class so it can be called from where we want to - eg. if we have to force an update of the JSON file, we can have the surface controller call the method directly, so we don't have to wait for the Hangfire job to run.

I should have some time over the weekend to finish the pull request ;)

### DIFF
--- a/OurUmbraco/Community/GitHub/Controllers/GitHubContributorController.cs
+++ b/OurUmbraco/Community/GitHub/Controllers/GitHubContributorController.cs
@@ -1,14 +1,9 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Net;
-using System.Text;
 using System.Web.Mvc;
-using Newtonsoft.Json;
-using Newtonsoft.Json.Linq;
 using OurUmbraco.Community.GitHub.Models;
 using OurUmbraco.Community.GitHub.Models.Cached;
-using RestSharp;
 using Umbraco.Core.Logging;
 using Umbraco.Web.Mvc;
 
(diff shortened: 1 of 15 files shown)


OurUmbraco/Community/GitHub/Controllers/GitHubContributorController.cs
```diff
@@ -121,7 +120,7 @@ public List<GitHubGlobalContributorModel> GetContributors()
 
             string[] login = System.IO.File.ReadAllLines(configPath).Where(x => x.Trim() != "").Distinct().ToArray();
 
-            var githubController = new GitHubController();
+            var githubController = new GitHubService();
```
bjarnef: Maybe rename this to `service` like in `ScheduleHangfireJobs.cs`

```
// Initialize a new service
GitHubService service = new GitHubService(); 
```

abjerner: I haven't updated the controller yet (besides what Resharper did when renaming the class). But yes, the variable name should reflect the new name of the class.

abjerner:
The controller and a bit more of the service class has now been updated, so everything now runs with the new logic.

@Shazwazza With the updated code, visitors will see a "Could not load recent GitHub contributors." error message if the cache file isn't there. Since the code now relies on Hangfire, the job should probably be executed manually through the Hangfire dashboard initially when the new code is deployed. Or through this URL: `/umbraco/Surface/GitHubContributor/GitHubGetContributorsResult?force=true` (which is public btw, so any can call it - not sure if that is a problem).

abjerner:
@Shazwazza @nul800sebastiaan Can we have a look at merging this?

The Contributions section is almost consistently showing the wrong numbers, as Niels removed some of the error checking, and the requests for one or more repositories fail when the cache is being populated (and the wrong stats are then being cached).

GitHub will invalidate their cache whenever a change happens in the master branch of a given repository. It seems `Umbraco-CMS` and `UmbracoDocs` are repositories most often the one left out when the stats are wrong - probably because they are also the two most active repos.

Whenever I visit Our and see the wrong stats, I do a force update (via the controller described above), so here is an example of how it looked today:

Just a few minutes ago:
image

After I ran a force update (which I had to run twice - due to the removed error checking):
image

I know it perhaps wasn't that clear whether this pull request is ready to be merged, as I still had "work in progress" in the title. I have updated the title, so it no longer says "work in progress" ;)

Shazwazza:
Yup we should defo get this in! @nul800sebastiaan could you do this today/tomorrow?

nul800sebastiaan:
Alright everyone, this is now all merged and deployed, let's see if it keeps up better this time! It still will refresh twice a day, changed it from daily. It ran, successfully, from Hangfire just now and @abjerner your score is now 27 instead of 19, so it looks like it got data from the OurUmbraco repo!

Great stuff!!

PR #123, opened 2017-07-13
Optimize github contributors for different breakpoints on frontpage

bjarnef:
Issue: http://issues.umbraco.org/issue/OUR-462

I have optimized the number of github contributors per row depending on different breakpoints.
Furthermore the white background is also aligned vertically with the other in e.g. forum posts and center the avatars, which I think look better e.g. when there is only two elements on last row.

**Before**
image

**After**
image

**Before**
image

**After**
image

It also fix these issues on mobile.
Same width of the white background as for forum posts, meetups etc.
screenshot_20170713-170647

Init number of github contributors fit to fill the row.
screenshot_20170713-170658

The smaller avatar in screenshot above is because the original image is too small for retina screens - the original size is only 200x200px https://avatars3.githubusercontent.com/u/1769475
Alternatively it can use ImageProcessor to handle size of the image, which allow upscale and allow remote domains like https://avatars3.githubusercontent.com

I think the  avatars can have image urls from avatars1.githubusercontent.com , avatars2.githubusercontent.com and avatars3.githubusercontent.com ... not sure if there are more.

### DIFF
--- a/OurUmbraco.Client/src/scss/sections/_contributors.scss
+++ b/OurUmbraco.Client/src/scss/sections/_contributors.scss
@@ -1,10 +1,7 @@
-﻿#github-contributors {
-
-    small.link-list {
-        font-style: italic;
-        font-weight: normal;
-        text-transform: none;
-    }
+﻿small.github-repo-list {
+    font-style: italic;
+    font-weight: normal !important;
+    text-transform: none !important;
 }
 
 .github-contributor-list {
--- a/OurUmbraco.Client/src/scss/sections/_contributors.scss
+++ b/OurUmbraco.Client/src/scss/sections/_contributors.scss
@@ -1,21 +1,52 @@
 ﻿#github-contributors {
+
+    small.link-list {
+        font-style: italic;
+        font-weight: normal;
+        text-transform: none;
+    }
+}
+
+.github-contributor-list {
+    justify-content: center !important;
     &:not(.all) {
         .contributor {
-            &:nth-child(n+17) {
-                display: none;
+            
+            /* Hide element no. 9 and above (2 items per row) */
+            @media (max-width: 440px) {
+                &:nth-child(n+9) {
+                    display: none;
+                }
             }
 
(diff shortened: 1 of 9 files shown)


emmagarland:
Nice one @bjarnef

bjarnef:
In the commit https://github.com/umbraco/OurUmbraco/pull/123/commits/b85a2813967c60281244285deb723e275a0c805e I have added ImageProcessor to handle sizing of the avatars, so it allow the image to upscale, so the overlay with additions and deletions has enough space.

E.g. this avatar is only 200x200px https://avatars4.githubusercontent.com/u/1769475?v=4&s=336 even it set size to 336px width, which is needed to ensure the images are sharp on devices with device pixel ratio 3, e.g. Samsung Galaxy Phones.

image

bjarnef:
Note that `/config/imageprocessor/cache.config` has this settings: 

```
<setting key="VirtualCachePath" value="~/../wwwcache/IPCache" />
```

I needed to create this folder manually, otherwise ImageProcessor returned a ysod saying that this folder didn't exists.

So I have create `/wwwcache/IPCache` folders manually, where `wwwcache` folder is at same level as `OurUmbraco`, `OurUmbraco.Client` and `OurUmbraco.Site` folders.

bjarnef:
I have noticed that the avatar image urls is different and might change from time to time. For now I have seen image urls using `avatars0` to `avatars7`.
https://github.com/bjarnef/OurUmbraco/blob/4d1948da539e61add8f27345594ef18904b77e29/OurUmbraco.Site/config/imageprocessor/security.config#L22-L29

I am not sure if you can use asterisk/wildcard or a pattern in these cases, e.g.
```
<whitelist>
     <add url="https://*.githubusercontent.com" />
</whitelist>
```
or

```
<whitelist>
     <add url="https://avatars*.githubusercontent.com" />
</whitelist>
```

nul800sebastiaan:
Looks good and works.. let's keep an eye on the avatars* then! :)

PR #124, opened 2017-07-22
OUR-464: Master badges on Our profiles

callumbwhyte:
I think it would be great for those who have achieved the "Master" status to have a badge on their Our profiles. We currently have the certification profiles on Umbraco.com but there's currently no indicator on Our right now.

This was added to the issue tracker here: http://issues.umbraco.org/issue/OUR-464

I have created a "Master" member group, along with the relevant styles to render the badge on profiles and also a migration to perform the necessary database changes.

I have also introduced a "Badges API" into the solution. The idea of this endpoint is to remove the need for manually assigning badges to Our profiles. Right now the API is setup to handle "Master" and "c-trib" badges, but the code has been written so this can easily be extended to more. Within the API you are also able to assign karma points to the member.

The API endpoints require a pre-shared token to be sent as the "token" header, which is to be matched against a value in appSettings. I couldn't find how to create new appSettings in the project, the following would need creating:

- "BadgesApiToken" <- the key sent in the "token" header to authenticate against, should be kept secret
- "CoreContribBadgeKarmaPoints" <- the number of karma points to be given when the "c-trib" badge is earned
- "MasterBadgeKarmaPoints" <- the number of karma points to be given when the "c-trib" badge is earned

### DIFF
--- a/OurUmbraco.Client/src/scss/core/_helpers.scss
+++ b/OurUmbraco.Client/src/scss/core/_helpers.scss
@@ -27,7 +27,7 @@
 	$color-roles-purple: #df89ff;
 	$color-roles-orange: #e8b142;
 	$color-roles-blue:	 #5dacff;
-    $color-roles-indigo: #5d00a2;
+	$color-roles-indigo: #5d00a2;
 
 	$color-headline: 	 rgba(#000, .8);
 	$color-text: 		 rgba(#000, 1);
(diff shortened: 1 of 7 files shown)


nul800sebastiaan:
Hey @callumbwhyte ! I really like the ideas of adding badges for masters etc. We're currently working on updates to get the contribution badges automated and it will use a pull system instead of your push system. Then we can also award some stuff retrospectively.

At some point the training levels will be available for us to pull in as well. 

This is all in progress so no need to change your stuff now, I'll close this but thanks for the efforts and enthusiasm! #h5yr  🍻 👍

callumbwhyte:
Glad you're working on the far superior pull based approach @nul800sebastiaan - this was my initial stance but obviously needs someone from HQ to do some magic on your side. Looking forward to seeing this in action, including "Master" badges, in the future!

PR #128, opened 2017-08-30
- removed h3 font-weight 100 style on desktop

prjseal:
Problem.
On the forum page, when viewing above min width, the h3 font goes blurry because the font weight is set at 100. But if you remove this rule it looks great at the other rule where it is set at 600. You can see this on the smaller width. 

Solution
I have removed the rule which changes the font-weight for the h3 to 100.
This allows it to be styled with font-weight 600 which looks so much better.

### DIFF
--- a/OurUmbraco.Client/src/scss/pages/_forum.scss
+++ b/OurUmbraco.Client/src/scss/pages/_forum.scss
@@ -179,10 +179,6 @@
 
 
 				@media (min-width: $md) {
-					h3 {
-						font-weight: 100;
-					}
-
 					span {
 						font-size: .9rem;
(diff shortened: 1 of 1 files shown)


Eyescream:
Great pull request! 👍 
Please merge this in, it's painful to read right now

prjseal:
I'm glad it's not just me.

nul800sebastiaan:
Thanks Paul! That helped a little bit but a lot of the text still looked very bad on Windows. I'm changing the font to Lato, which is what we use on other Umbraco sites as well, so it will be consistent and we know the font behaves much better on all sizes!

PR #132, opened 2017-10-05
Form method changed to pass form values as querystring params

mzajkowski:
I noticed that versions during comparison on https://our.umbraco.org/contribute/releases/compare page are not passed as a query string parameters and the result cannot be easily shared between people. 

I thought the whole feature is missing, but I noticed everything is there, but the wrong form method was causing those params to dissapear. 

Changing it from POST to GET fixed the problem. Hope it will help others too :)


nul800sebastiaan:
Cool! Not sure why we never did this, I think I thought of changing it as soon as we got this feature. 

Anyway, great job, thanks! It will be deployed soon!

mzajkowski:
Yup, trivial and I just spotted it as I was in need to share some results from this page. I'm also pretty sure it was working fine at some moment in the past. No matter what - it's nice to help! Cheers.

PR #140, opened 2017-12-23
Added footer link to the Terms of Conditions

abjerner:
As suggested by Niels here: https://our.umbraco.org/forum/using-umbraco-and-getting-started/89827-impact-of-slack-community-on-these-forums-and-on-the-umbraco-community-in-general#comment-283642

### DIFF
--- a/OurUmbraco.Site/Views/Master.cshtml
+++ b/OurUmbraco.Site/Views/Master.cshtml
@@ -168,7 +168,7 @@
                         }
                         <br />
                         <br />
-                        <a href="/terms-and-conditions">Terms of Conditions</a></div>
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Terms *of* service ! 😉

nul800sebastiaan:
Or, Terms **and** conditions I guess is what we called it.

abjerner:
Bah. Seems I was a bit too quick. I've made another commit so it's now called **Terms and Conditions** as the page ;)

And yes, I think a page like that is normally called Terms of Service rather than Terms and Conditions.

nul800sebastiaan:
Thanks mate, some better pages for this are in the works and we can link them up when they're done (in the next few weeks). I'll close this one.

PR #144, opened 2018-01-30
OUR-481

Tidmand:
OUR-481 updates to css to handle break on long forum post titles and post categories


nul800sebastiaan:
Sweet, thank you!

PR #145, opened 2018-01-30
OUR-109

Tidmand:
OUR-109 added styling to defined actionButton class to avoid overflow caused by mega size buttons. It looks like a lot of changes in the scss file, but it's actually only a group of lines inserted at line 166.


nul800sebastiaan:
Yeah, you can filter whitespace changes by adding `?w=1` to the URL : https://github.com/umbraco/OurUmbraco/pull/145/files?w=1


Makes it easier to see the real differences! I'll have a look at it this week, thanks!

PR #151, opened 2018-02-14
WIP - New "Videos" section

abjerner:
- [x] Logic for working with the Vimeo API<br /><br />
- [x] Content type, template and content item for "Videos" (including migration)<br /><br />
- [x] View and partial view to list videos by their category<br /><br />
- [x] **Responsive**<br />Our is responsive, so the new videos list should also be responsive<br /><br />
- [x] **Embed**<br />Videos currently link to Vimeo so they can be viewed there. The list should support playing videos directly from the list.<br /><br />
- [x] **Query string -> folders**<br />The URLs should use folders rather than being based on the query string. Could be solved with URL rewriting.<br /><br />
- [x] **Hangfire**<br />We need to create a Hangfire job to update the videos on a regular basis.<br /><br />
- [x] **Singular/plural**<br />Various counts should support singular amount - eg. *1 video* and *1 play*.<br /><br />
- [ ] **Cleanup the CSS**<br />The CSS currently lives in the view as I was playing around. This needs to be moved to the SCSS files.


nul800sebastiaan:
We can clean up CSS later, I'll merge it now so we can test it out on live. Awesome stuff so far!!

PR #155, opened 2018-02-19
Videos: Added "All videos" menu item

abjerner:
I think this makes sense. Eg. to have a menu item that is selected by default.

The menu item doesn't listed the individual years as the logic currently doesn't support browsing those without a category - but should this be supported as well?

image


nul800sebastiaan:
👍

PR #156, opened 2018-02-19
The CSS for the new videos section has now been moved out of the view to it's own SCSS file

abjerner:
This also means that we can reuse the CSS for the community videos page :D


nul800sebastiaan:
Wonderful!

PR #159, opened 2018-02-22
UDUF: OUR-487 - Adding flag to set projects as retired

callumbwhyte:
This change adds 2 new fields to projects (packages) on Our whereby you can set a package to be "retired" and supply a message explaining the reason (e.g. it is superseded by a new package). These settings can be managed through the package editor.

When a project is set to "retired" a box will appear at the top of the project page with a default message - this message can be overridden with the "retiredMessage" field.

I guess this is a starting point for a more fully featured implementation and way of handling retirement / deprecated packages as per the discussion on the issue tracker (http://issues.umbraco.org/issue/OUR-487) and Twitter (https://twitter.com/KevinGiszewski/status/958340250237063168)

Feedback welcome!


nul800sebastiaan:
Good stuff Callum, thanks! You could've tested that it actually worked ;-)
But that was fixed with a simple update. I've also removed the retired packages from Examine so it won't show up in search results any more.

PR #160, opened 2018-02-22
OUR-500 fix

villian:
very very small fix )


nul800sebastiaan:
Thanks mate! Not the correct fix.. but fear not, I've just fixed it, your MVP badge is coming back soon!

PR #168, opened 2018-03-05
Switch documentation Markdown conversion to use Markdig 

Jeavon:
Switch documentation Markdown conversion to use Markdig so that it supports all things GitHub such as tables and GFM code blocks

### DIFF
--- a/OurUmbraco.Client/src/scss/pages/_documentation.scss
+++ b/OurUmbraco.Client/src/scss/pages/_documentation.scss
@@ -362,7 +362,7 @@ $color-overview: lighten($color-our, 20%);
 
 
 
-    .col-xs-6 {
+    .col-xs-6, .col-sm-6 {
         min-height: 133px;
 
 
--- a/OurUmbraco.Site/OurUmbraco.Site.csproj
+++ b/OurUmbraco.Site/OurUmbraco.Site.csproj
@@ -126,6 +126,9 @@
       <HintPath>..\packages\Lucene.Net.2.9.4.1\lib\net40\Lucene.Net.dll</HintPath>
       <Private>True</Private>
     </Reference>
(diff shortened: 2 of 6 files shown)


Jeavon:
Instead of image
We get 
image
When someone sends us a Markdown table

dampee:
This would be so much easier to write documentation on github.  The preview on github hardly every matches the final version.  Great pull request!

warrenbuckley:
@nul800sebastiaan do we use Markdownsharp anywhere else in the site/solution?
If so can I recommend we try to swap over to using this lib - so there is one consistent way of markdown being parsed.

Jeavon:
@warrenbuckley I think it's used for Forum comments (ForumExtensions.cs)

warrenbuckley:
@Jeavon hmm I would like to say that we could swap it out & it would be a relatively safe swap, but @nul800sebastiaan has far greater knowledge of the codebase & if there is any specifics or quirks we use from MarkdownSharp.

Probably won't hurt to have both - but just be nice to be clean & tidy & stick with one - @nul800sebastiaan will let you test, review & get this one merged as this is your baby :)

Jeavon:
I agree @warrenbuckley, I was already going to try to swap this in for the forum comments too (probably without all of the extensions enabled), but initially we really need it for the documentation so we can fix the formatting

warrenbuckley:
OK if it makes life for the Docs team easier then it makes sense to run with both lib's to begin with so it can be reviewed & rolled out sooner & then we can do another pass/iteration on this at a later date to have one lib.

Nicholas-Westby:
You might want to take a look at this: https://twitter.com/MrMarsRed/status/973659455597428737

Basically, I noticed recently that URL's I paste into forum replies no longer automatically get converted into clickable links (i.e., I have to manually link them). It was brought to my attention that this pull request may be related to that issue.

Jeavon:
@Nicholas-Westby the change to Markdig only affects documentation currently.

PR #169, opened 2018-03-05
Google Map - Display members around the world

warrenbuckley:
Adds a PR to bring back the old member location Google Map - users have to opt in by giving their lat & lon on their member profile. So it's totally optional.

* Adds Examine OnGatheringNodeData event - to add new field to index - used to help filter/find members who have lat/lon info on their profile
* Adds an API Controller - to query Examine for members who have Lat & Lon fields set
* Simple partial view that uses CDF to require the JS it needs
* External JS file for loading our markers from API endpoint & the marker info window
* Adds Google's community library for clustering markers

**Note**
We need to test how well this performs with more/real world data than in the test DB.

### DIFF
--- a/OurUmbraco.Site/Assets/js/lib/markerclusterer.js
+++ b/OurUmbraco.Site/Assets/js/lib/markerclusterer.js
@@ -1,0 +1,1288 @@
+// ==ClosureCompiler==
+// @compilation_level ADVANCED_OPTIMIZATIONS
+// @externs_url http://closure-compiler.googlecode.com/svn/trunk/contrib/externs/maps/google_maps_api_v3_3.js
+// ==/ClosureCompiler==
+
+/**
+ * @name MarkerClusterer for Google Maps v3
+ * @version version 1.0.1
+ * @author Luke Mahe
+ * @fileoverview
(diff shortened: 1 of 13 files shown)


warrenbuckley:
Still todo:
- [ ] Filter/query the API endpoint based on the current viewport of the map
- [ ] We may need to look into GeoSpatial for the index to utilise the lat & lon
- [ ] By adding the avatar to the marker - its trying to load each markers avatar 40+MB OUCH!

warrenbuckley:
@nul800sebastiaan it's ready for a re-review mate & should be quite snappy now as we only get a subset of data for the current map view, unless you zoom all the way out :-P

warrenbuckley:
- [ ] Move Google Maps API Key to an AppSetting
- [ ] Display a list/table of members all at exact same location - such as London with 100+
- [ ] Sort the list - by karma (so most active is display first at the location)
- [ ] Add copy about how to be added or removed from the map (by updating profile)
- [ ] Add the fields back to the edit profile page (Lat & Lon) & use Google Map
- [ ] Add copy note to the edit profile about being sensible and too not share exact locations
- [ ] Add a way to clear/remove lat/lon that is set on a member profile

PR #178, opened 2018-03-11
Correcting spelling of UmbraCoffee in YouTubePlaylists.json

callumbwhyte:
This isn't a big change but I guess every little helps, right? ;-)

I noticed "UmbraCoffee" was spelt wrong (UmbraCofee) in the new YouTubePlaylists.json so I've fixed it...


nul800sebastiaan:
Duh!
Thanks, indeed, every little thing helps!

PR #184, opened 2018-03-18
WIP - Community calendar

### LINKED ISSUE #182: Community section: create calendar of Umbraco related events
This includes:

  - Festivals
  - Meetups
  - Things like CodeCabinn
  - Niels is speaking at a conference
  - etc

Should be able to filter by type so we can list only festivals or only meetups.

We should be able to do some content management on festivals from the backoffice to provide a description, link, maybe a banner.

RSS feed of events would be an added bonus.

We can re-use the design from the roadmap so we have a timeline with events on it.

abjerner:
closes #182

- [x] Implement migration for adding the community **Calendar** page
- [x] Load meetups using the existing logic for Meetup.com
- [x] Loads festivals from content cache (instead of being harcoded)
- [x] Lists years in the sidebar (works as a filter)
- [x] New sidebar HTML and CSS for the timeline
- [ ] Move timeline CSS from partial view to it's own `SCSS` file


nul800sebastiaan:
😮😎

abjerner:
@nul800sebastiaan I've now pushed some more changes - one of them being able to filter by year. Default is upcoming events, which are sorted in descending order. When a specific year is selected, the events are instead sorted in ascending order.

The meetups are loading using the existing logic for Meetup.com. This works fine, but was designed to show upcoming meetups, so we probably need to update the logic to include all meetup events of a each Meetup.com group. But perhaps this is a step for Community Section v2?

Other events are currently hardcoded. As we discussed, these should come from the content section within the Our Umbraco site. Could this be a task for you (or the HQ in general), as you probably have a better idea on what information should be added?

I'm still not entirely sure about the design, as the descriptions seems to make the list appear more "messy". Also, meetups have descriptions in different forms - and I noticed that the German meetups have German descriptions (VERDAMMT).

Each type of meetup should probably also have their own icon, which could be shown in the green circle of the corresponding timeline item. I haven't look into whether there are existing icomoon icons that we could use for this.

My idea was also that a festival gets a bigger circle in the timeline (inspired by the releases page). Festivals could also have even bigger focus - and show more information like link to website, where to buy ticket etc.

nul800sebastiaan:
Updated with the latest and made a new DocType for creating calendar items.

image

nul800sebastiaan:
Alright, got items from the content cache now, UDUF here is coming from a content item in Umbraco, the rest from the existing meetup cache. Also added an icon for each of them. It's ugly but it works.

image

nul800sebastiaan:
Merging to master now so it can be deployed and some content can be added - feature is hidden for most people, except for admins.

PR #185, opened 2018-03-19
Fixes twitter on homepage if keys are invalid

warrenbuckley:
Was expecting twitter to throw some kind of unauth ex, but we get a null. Now we return an empty array so the view doesn't blow up/throw a 500

Done as this was annoying me on local debug copy, when homepage fired up it would throw due to a null trying to be iterated on in the view


warrenbuckley:
Done as a PR as I wanted to test GitHib was happy again with my GPG verified commits again (after clean wipe)

PR #187, opened 2018-03-21
Open "Join the discussion on Gitter" in a new tab/window

abjerner:
As the link is external, it should probably open in a new tab/window


nul800sebastiaan:
Yep, good stuff, thanks!

PR #188, opened 2018-03-21
Be less of a creep!

warrenbuckley:
Removes GitHub, Twitter & Karma count from the marker Pins & API response. We still sort members by karma though for many users at location when zoom level is close.


warrenbuckley:
Creep

warrenbuckley:
Now with your requested changes @nul800sebastiaan 
Creeping

PR #190, opened 2018-03-22
Fix Gitter Bootup issue - if room no longer exists

warrenbuckley:
Better error handling for trying to connect to rooms that do not exist


warrenbuckley:
@nul800sebastiaan Check & test this works OK, but should fix up the issue

PR #191, opened 2018-03-22
Radar - bleep bloop beep

warrenbuckley:
GeoCities Radar

* Adds 3rd Party Hangfire.Console library - https://github.com/pieceofsummer/Hangfire.Console
* New Examine Field added to indicate if the field 'location' exists & has a value set (allows us to find these members with Examine query)
* Adds new Hangfire Job that queries Examine for members missing lat/lon but has legacy free textfield Location field & attempts to look them up with Google Geo API

### DIFF
--- a/OurUmbraco.Site/Views/Partials/Community/Map.cshtml
+++ b/OurUmbraco.Site/Views/Partials/Community/Map.cshtml
@@ -1,7 +1,8 @@
-﻿@using ClientDependency.Core.Mvc
+﻿@using System.Configuration
+@using ClientDependency.Core.Mvc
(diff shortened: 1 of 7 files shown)


warrenbuckley:
## Note
As we add a new field `hasLocationTextFieldSet` to the member index. The members index needs to be reindexed so that the query to find these members can be run/found in the Hangfire Job

warrenbuckley:
@nul800sebastiaan all ready for a code review & test (ensure you test locally & on Azure test env)
Just as don't wanna have missed something/mess something up with the live member data

PR #203, opened 2018-03-27
Updated codeshare logo

prjseal:


warrenbuckley:
Thanks @prjseal 
All merged in, @nul800sebastiaan will organise his magic for when it will get deployed

PR #228, opened 2018-04-11
Added dummy file so the /media/avatar folder exists in the repo

abjerner:
Generating new fake avatars fails without the folder present (generic GDI+ exception)


nul800sebastiaan:
Hehe, sure why not! Could have made some code to create the dir if it doesn't exist but this is a nice hack ;-)

PR #234, opened 2018-05-03
WebForms? Yikes!

abjerner:
The icon to the left in each result result item isn't rendered properly since the statement outputting the class name hasn't been converted from the old WebForms.

With this commit, I've changed the line so it uses the proper Razor syntax, and the results then look like as shown below:

image


warrenbuckley:
Ace thanks for the fix, good find

PR #236, opened 2018-05-20
Documentation Tag support with YAML

dampee:
This PR will make it possible to add YAML on top of every documentation page


abjerner:
A little side question, as this could be influenced by the YAML tags: should each page show information about the areas that they describe?

Say I'm searching for Models Builder, the search could then show the area(s) in which the page is located.

image

The area could be in a number of different ways:

- Based on an "area" tag in YAML
- Automatically on the folder structure (eg. the titles of the two or three topmost pages)

Or we could just show the tags of the page instead (not sure whether this is a part of your pull request).

dampee:
Hi @abjerner 
Thanks for your thoughts!
Not sure if we would be able to extract it from the folder.  Certainly as that might change over time.
Using a YAML tag would be feasable, but that would mean we will need to update all documents.  Which is not impossible at all.  

Personally I would use the YAML tags.  I can only dream of a situation where they enable us to navigate across documentation not by folder but by e.g. task, related topic, ...

dampee:
The only things to verify:

- [ ]  the current document not recognized [in this line](https://github.com/umbraco/OurUmbraco/pull/236/files#diff-cad945bc4df172d73846e991a04c901dR66).  If it is recognized, it will don't add a link in the view.
- [ ] test if all different version scenario's are correct (using versions from, to, both or none in YAML)
- [ ]  consider if we need to do something with the ordering of the version documents
- [ ]  tweak the keywords boost (ok for now i guess). -- let's get it in first and tweak afterwards.

Do you think of something else?

dampee:
Did a rebase to another branch and created a new pull request: #308

Issue #227, opened 2018-04-10
Access member profiles by nickname


greystate:
It has always bugged me that an Our member's profile is only accessible using their `id`.

I always have to do a search just to find Janae's or Fuji's () profile - heck, I can't even remember my own id :-)

Wouldn't it be possible to create a **ContentFinder** that could find the right member, given their GitHub user or Twitter handle? E.g. for me, they're the same, but `https://our.umbraco.org/member/greystate`  would work anyway.

I can see two problems:

1. It _is possible_ for someone to have the same GitHub username as someone else's Twitter handle but that should be very few, but even then, we could maybe show a list of the two?

2. Some members have neither a GitHub user nor a Twitter handle - but their id-URL would still work then...

Would there be any security-related issues with this?

abjerner:
@greystate I think this would be awesome.

I looked a bit at the code, and we can update the existing macro partial (yes, the public profile is a macro) along with a small rewrite rule. `Web.config` already has:

```xml
<rule name="MemberProfile">
  <match url="member/(\d+)" />
  <action type="Rewrite" url="member/?id={R:1}" />
</rule>
```

so we can add:

```xml
<rule name="MemberProfileUsername">
  <match url="member/(\w+)" />
  <action type="Rewrite" url="member/?username={R:1}" />
</rule>
```

If we use the GitHub username, the code for finding the member could look something like:

```C#
// Get a reference to the member searcher
BaseSearchProvider searcher = ExamineManager.Instance.SearchProviderCollection[Constants.Examine.InternalMemberSearcher];

// Initialize new search criteria for the GitHub username
ISearchCriteria criteria = searcher.CreateSearchCriteria();
criteria = criteria.RawQuery(String.Format("github:{0}", username));

// Make the search in examine
ISearchResults results = searcher.Search(criteria);

// Get the member based on the ID from the search result
if (results.TotalItemCount > 0)
{
    member = Members.GetById(results.First().Id);
}
```

I tested this on my local Our installation, and it works like a charm. But it wont work if the GitHub username is numeric (not sure if GitHub supports that), as the Our numeric ID rewrite rule will be matched first.

I think we should either stick entirely with the GitHub username (not Twitter) or let members claim their username independently of GitHub and Twitter.

@nul800sebastiaan Any thoughts on this? I be happy to make a pull request with my code for the GitHub username.

[...]

PR #237, opened 2018-05-21
GitHub usernames (vanity URLs)

abjerner:
A follow up on @greystate's #227 and other discussions related to GitHub linking.

With this pull request, it is no longer possible to specify a GitHub or Twitter username directly. Instead, members can now link with their respective GitHub and Twitter accounts.

The linking is handled via OAuth, we members now prove that they accounts actually belong to them. And spammers can't just specify a URL as their username.

Linking now lives under the *Edit profile* page. By default, this page would look something like shown below:

image

In this case, the member hasn't yet linked with their GitHub or Twitter accounts. Notice that the path of the URL is now `/members/id:147279/` (opposed to `/member/147279` as it is right now on the live site).

For an member profile that already has been linked with GitHub and Twitter, the page will look like:

image

As I've linked GitHub with my test account, the member profile URL is now `/members/abjerner/` instead.

This pull request includes the necessary rewriting rules for making this happen. The old member URLs will now redirect to the new ID based member URL, but not to the username based URL.

**GitHub OAuth**

As Our doesn't appear to be using an existing GitHub app, one should be create via the following page:

https://github.com/settings/developers

The callback URL should be something like:

https://our.umbraco.org/umbraco/surface/Profile/LinkGitHub

**Twitter OAuth**
Twitter consumer key and consumer secret is already specified in Web.config, so these should be good. But a callback URL should be specified for that particular app, which can be done here:

https://apps.twitter.com/

Also notice that the permissions to *Read only* as we don't want to require more information that we need.


greystate:
*Praise Be!*

This is all kinds of awesome - major kudos to you @abjerner 💯 🥇

abjerner:
**Username**
I can't remember who it was, but someone at Codegarden suggested that it should perhaps still be possible to specify a username without having a GitHub account.

What's the HQ stance on this? Restricting the username to be just members with a GitHub account may exclude some members. But then again, using just GitHub makes sure that usernames are unique, and "less friendly" members aren't able to "steal" the username of other members.

Also, I think that most members who would add a username (or at least getting value from doing so) might already be contributing to Umbraco via GitHub, and thus already have a GitHub account. But this is just my opinion - it have no data saying one or the other 😮  

**UI**
As we were various groups at the retreat working on Our, some of the work collides a bit involving the **Edit profile** page. So perhaps we (or more likely someone at the HQ) should look at this page with a broader perspective. I know the wire-frames aren't online yet, but just wanted to add a note here ;)

abjerner:
Wow - that was a lot of commits 😄 

Btw - this might go for a future PR, but what kind of data should we store when linking with GitHub and Twitter? Right now it's just the username and nothing else. But we might need other information for other features - eg. using the ID when attributing the member would technically be a bit safer as username can be changed.

nul800sebastiaan:
Haha, I only did a merge! 🙈 

Yeah, definitely might be nicer to store the ID indeed! Not sure what other data we get?

abjerner:
@nul800sebastiaan something like this:

https://api.github.com/users/abjerner

```
{
  "login": "abjerner",
  "id": 3634580,
  "node_id": "MDQ6VXNlcjM2MzQ1ODA=",
  "avatar_url": "https://avatars0.githubusercontent.com/u/3634580?v=4",
  "gravatar_id": "",
  "url": "https://api.github.com/users/abjerner",
  "html_url": "https://github.com/abjerner",
  "followers_url": "https://api.github.com/users/abjerner/followers",
  "following_url": "https://api.github.com/users/abjerner/following{/other_user}",
  "gists_url": "https://api.github.com/users/abjerner/gists{/gist_id}",
  "starred_url": "https://api.github.com/users/abjerner/starred{/owner}{/repo}",
  "subscriptions_url": "https://api.github.com/users/abjerner/subscriptions",
  "organizations_url": "https://api.github.com/users/abjerner/orgs",
  "repos_url": "https://api.github.com/users/abjerner/repos",
  "events_url": "https://api.github.com/users/abjerner/events{/privacy}",
  "received_events_url": "https://api.github.com/users/abjerner/received_events",
  "type": "User",
  "site_admin": false,
  "name": "Anders Bjerner",
  "company": "Skybrud.dk",
  "blog": "",
  "location": "Horsens, Denmark",
  "email": null,
  "hireable": null,
  "bio": "Umbraco MVP - System developer at Skybrud.dk (@skybrud)",
  "public_repos": 61,
  "public_gists": 27,
  "followers": 29,
  "following": 14,
  "created_at": "2013-02-19T08:09:09Z",
  "updated_at": "2018-05-20T12:45:15Z"
  }
```

If we actually do need the email address of the (authenticated) user, we can also get that.

nul800sebastiaan:
You know what, it might actually be very cool for people to be able to sign up with their github account - though we do also need to work on "Umbraco Id" where your account is valid for Our, umbraco.com, TV, etc.
But I guess as long as that doesn't exist we could let people sign up with GH, that adds a big level of trust, we might not have to manually approve people before they start posting. It might also make it much easier to sign up etc..

Hmmm, let's not build it yet, but I'll think about it some more!

abjerner:
That could totally be possible. Someone also suggested that at Codegarden.

I don't know how the OWIN providers work, but I know we can do it with my package.

abjerner:
Anyways, before you merge this one, I can add another property on members to store the GitHub user data (should we save Twitter user data as well)?

What data should we then store? Just everything GitHub/Twitter throws at us, or just stuff like id, username and name? It's data publicly available via the API anyways.

abjerner:
I worked a bit more on this, and I've ended up adding two new properties for GitHub - a `githubId` property for just storing the GitHub user ID. This is then automatically indexed in Examine so we can use it for lookups instead of the username. I'm not sure whether we actually need it, but the JSON object received from GitHub is now also stored in a `githubData` property. I've added migrations so both properties are created automatically.

I've added similar properties for Twitter, so there's now a `twitterId` property and a `twitterData` property, although we don't actually need those at the moment.

nul800sebastiaan:
🎉

nul800sebastiaan:
One note: if you've filled in your github username before, it looks like it's already linked, but the vanity URL doesn't work until you unlink/relink it.

We should be able to migrate existing users at some point in the future though. 👍

PR #243, opened 2018-05-27
Improvements for the documentation edit button

ed-parry:
This is off the back of a discussion that we had during the V8 Documentation open space session, with the aim being to make it more obvious that this button can be used to edit the page contents. 

ourumbraco-docs-edit-button

Just a couple of small changes as I'm conscious to not move away from the standard styling. The GitHub icon is switched out (and removed) for the more typical pencil/edit icon, which may mean more and immediately signal to users that they can edit -- maybe the GitHub icon puts some people off, when it's actually very easy to make the changes and submit a PR (most of it is automated).

Also updated the text to try and make it more obvious the page contents can be changed. I played around with adding a hover/`title` but it isn't really the right use of the UX.

Just a first pass at this stage, but may help more people engage with the docs. Happy to update/revise as we see fit :)


ed-parry:
Also a quick note - I was slightly thrown by the Site/Client projects (still too tired after CG!) so if I've missed anything in either just shout. I'll double check again later on too.

ed-parry:
@Jeavon @OwainWilliams @tcmorris Pinging you on this one (trying to find others but people don't keep Twitter/Github users the same! 😅) as we discussed it during the V8 docs session so would value your input :)

OwainWilliams:
I still think it needs to be more obvious - not sure how yet but I'll put my thinking hat on :)

ed-parry:
Awesome, thanks @OwainWilliams. Thinking about it, maybe using the green CTA styling elsewhere on Our could work (the go-to buttons on the [home](https://our.umbraco.org/))?

OwainWilliams:
I think the green CTA could be an option - it keeps it in keeping and it's only  distraction when you are at the top of the page. Better to catch peoples eye than not see it at all.

ed-parry:
screen shot 2018-05-27 at 22 09 23

Mocked it up for reference. Kept the font/icon style for now rather than the larger text and arrow style on the full CTAs. 

Seeing this version looks more purposeful, and I agree @OwainWilliams - we'd rather everyone see it at least once and be aware. Regular users soon tune these things out if they're not for them.

OwainWilliams:
I think that looks more in keeping with the style of the site but also catches your eye, in a good way. 👍

OwainWilliams:
I've just realised actually - I think this PR should be within the Docs repo rather than the Our repo. Maybe worth re-creating an issue over on https://github.com/umbraco/UmbracoDocs/issues

nul800sebastiaan:
No, it needs to change in this repo so issue should be here. :-)

Would it be better to place the button the right side of the `<h1>`?  Now it kinda gets lost in the navigation, feels like it would get less attention above the search field.

OwainWilliams:
I think down next to the `h1` could be a good idea as you aren't actually editing above that line. 

The reason I was suggesting moving it to the docs section was because the chat about this button came up in the open space chat and I thought some people might be missing it in here, but makes sense to leave it here now it's been explained to me 👍 Cheers @nul800sebastiaan

abjerner:
A quick side question, but still somewhat related: Should the button link directly to edit mode on GitHub?

Right now the button will link to view mode - eg:

https://github.com/umbraco/UmbracoDocs/blob/master/index.md

But it is possible to link directly to edit mode:

https://github.com/umbraco/UmbracoDocs/edit/master/index.md

hartvig:
While we all want people to edit docs, do remember - in the mist of all enthusiasm - that the primary goal of a page in the docs is to be documentation. In other words, it should be obvious that you can edit the page but it shouldn't be the primary call to action.

@abjerner I believe linking directly to edit mode is a great idea

OwainWilliams:
It's finding a balance. 
The chat that came up in the open space chat was people a) didn't know they could help update the documentation and b) had never noticed the edit button.

skttl:
How about a bottom section like "Was this article helpful? If not _raise an issue_ or _edit the article_ to make it better".

That way you don't have to rely on the readers to scroll up again, in order to be aware of the edit possibility. And you add a way to get feedback on the article.

OwainWilliams:
@skttl this was also suggested during the talks and I think it was something that was hoped to be implemented. Fixing the 'edit' button was a quick fix before spending time on a 'Was this article helpful' option. I think that was going to be looked in to as part of a wider 'Our' revamp.

hartvig:
@OwainWilliams Indeed - I think the updated copy and the removal of the Github icon is fantastic.

ed-parry:
Thanks all - great suggestions and discussion.

I've mocked up how it looks with the button alongside the page title, and it feels a little crowded to me. I agree that it could be missed alongside the navigation, but that very top-right corner spot feels a strong location for engagement. Alongside the title it feels as though it takes away from the documentation itself? If these pages were more of a wiki style, I'd agree with the button being so close to the content title, as though the page would switch into an edit mode in-line, but currently the button in the top-right corner feels cleaner to me? 

screen shot 2018-05-28 at 10 50 56

What do we think?

ed-parry:
@abjerner Thanks, Anders - updated to use the edit link :)

tcmorris:
@ed-parry looks good. I think this a step in the right direction and is more apparent as to the intention. Placing this within the context of the doc is better too. If you were to place it next to the title, then I feel like it would need to be styled as a link and then it could easily be missed as you scan the page.

tcmorris:
@skttl I've created a separate issue for the 'was this page helpful' scenario: #246

Jeavon:
@tcmorris good thinking but don't worry about the upcoming version bits, we will incorporate/merge this into our work on that

ed-parry:
Cool, thanks all! I'll make a new commit this evening to bring the colour and position into this and then we're likely ready to go.

@tcmorris was your suggestion to leave the button in the top-right or put it alongside the heading? My personal preference is the top right, but happy to try either, especially with further changes in the pipeline.

tcmorris:
@ed-parry I think top right is fine. nice work!

ed-parry:
Updated and potentially ready to go unless anyone has any further recommendations or thoughts?

nul800sebastiaan:
Wow, that's a long discussion for a new button! ;-)

All good though, good to see the passion and the back and forth. The final solution looks great. It will be live in the next few days!

Thanks @ed-parry for the work and everyone else for the help figuring it all out!

ed-parry:
Thanks @nul800sebastiaan - guess that's what happens when you open issues in the immediate aftermath of Codegarden! Now onto the next one to keep the progress up. Cheers all!

PR #249, opened 2018-06-04
Separate contributing file

ed-parry:
Done the same for the UmbracoDocs project ([merged PR](https://github.com/umbraco/UmbracoDocs/pull/848))

This looks to be the recommended approach from Github, and the separate file is automatically pushed in front of users who create issues or pull requests. I've also updated the main README to remove the duplicated guidelines, but kept the heading and provided a new link.

Core already had this in place so bringing Our and Docs up to the same level.


nul800sebastiaan:
Thanks @ed-parry !!

PR #250, opened 2018-06-07
Update deprecated UserType

bjarnef:
`UserType` property is deprecated, but we can instead access `Groups ` property and check of any of these match "admin".


nul800sebastiaan:
Thanks @bjarnef - works like a charm (well apart from [the missing using statement](https://github.com/umbraco/OurUmbraco/commit/9ee9ccf41aad7a4145ebc2212add7283b327945a#diff-169b2b29ece2e69a6793fed748dbae3f) ;-) )

bjarnef:
@nul800sebastiaan sorry about that. I admit I didn't test it on Our project, but similar filter for Hangfire on another project :)

PR #258, opened 2018-06-13
Some Google Analytics changes

JeffreyPerplex:
Swapped the "Send pageview" and "Set dimension" statement. You should always first set the dimension and then send the pageview. I did it the other way around.


nul800sebastiaan:
👍 Will deploy in the next hour.

Issue #256, opened 2018-06-08
Breadcrumb menu link - wrong


syska:
When I'm on: https://our.umbraco.org/search?q=Test&cat=forum

The "Search" link in the "breadcrumb menu" links to the same page, while the page I'm on is still in the "breadcrumb menu " as "test".

Also reported here: https://our.umbraco.org/forum/contributing-to-umbraco-cms/92407-breadcrumb-link-wrong

jamiepollock:
@syska I've submitted a PR to fix this however as mentioned in #263 the default search page isn't much use to a user currently.

umbrabot:
Hiya @syska,

Just wanted to let you know that we noticed that this issue got a bit stale and might not be relevant any more.

We will close this issue for now but we're happy to open it up again if you think it's still relevant (for example: it's a feature request that's not yet implemented, or it's a bug that's not yet been fixed). 

To open it this issue up again, you can write `@umbrabot still relevant` in  a new comment as the first line. It would be *super* helpful for us if on the next line you could let us know why you think it's still relevant.

For example:

>@umbrabot still relevant
>This bug can still be reproduced in version 8.9.0

This will reopen the issue in the next few hours.

Thanks, from your friendly Umbraco GitHub bot :robot: :slightly_smiling_face:

PR #263, opened 2018-06-20
Fixes issue #256: "Search" breadcrumb wrong URL

jamiepollock:
The URL should take a user back the default search page not the current search term results.

--

Upon investigation this PR by itself isn't much use. The default search page does not have a search box for a user to interact with the page.

Perhaps a great discussion is required?

### DIFF
--- a/OurUmbraco.Site/Views/Partials/SearchResults.cshtml
+++ b/OurUmbraco.Site/Views/Partials/SearchResults.cshtml
@@ -9,7 +9,7 @@
             <div class="utilities">
                 <ul id="breadcrumb">
                     <li><a href="/" rel="nofollow">Our Umbraco</a></li>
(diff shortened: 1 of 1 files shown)


jamiepollock:
Fantastic. Thanks @dampee! Looks like we might be able to merge this in once that work is done then, eh?

nul800sebastiaan:
Thanks guys, I'll close this one then 😄

jamiepollock:
@nul800sebastiaan I'm not entirely sure #236 covers the same thing as this PR.

This is a short and sweet PR which fixes the search breadcrumb. Where as #236 is a overarching improvements to search in general.

I think unless @dampee, @Jeavon or marcemarc can say otherwise there is still some merit to this. :)

Issue #260, opened 2018-06-16
Feature suggestion: Option to edit post immediately after it has been created


BatJan:
I noticed that whenever one creates a new post the context menu is limited to only have 2 of the usual options. One is delete and the other one I think is "Copy link".

However I find that I make many typos when I write so after posting I need to edit the post. But currently one will need to refresh the page using ctrl+f5 in order to see the full range of options on the post where it's then possible to edit instead of being able to do it immediately.

Does this make sense? :-) Don't know if it's at all possible though. But would be nice though.

I'll be happy to have a look at it myself when I get some spare time and figure out.

greystate:
Yep - I’ve actually seen many complain about not being able to edit their post (because they hadn’t refreshed).

nul800sebastiaan:
Oops, wrong button! 🙈 

I remember we had the same problem when editing a comment, so I added a refresh at some point after submitting the edit. It's not very sexy but that would be the only easy way to make this work, only after the refresh do we have the id of the post available to be able to show the appropriate buttons. 

Of course the ideal solution would be to send the correct data back after posting and populate the buttons that way.

jamiepollock:
@BatJan @nul800sebastiaan hey guys, I have a fix in place for this. :)

nul800sebastiaan:
Resolved with #266

PR #266, opened 2018-06-21
Fixes Issue #260 edit new posts

jamiepollock:
Hey folks,
I've done some work so that a logged in comment owner or admin can manage (read: edit or delete) a post upon creation.

This is from issue #260. Eventually we could extend this to have the topic owner involved to "Mark as Solution" but that feels outside of the scope of the PR.

Thanks,
Jamie


nul800sebastiaan:
Makes sense, thanks @jamiepollock !

PR #268, opened 2018-06-26
Updated master template and footer links to the new .com domain

ed-parry:
Super small update as I noticed the site footer still referenced our.umbraco.org rather than our.umbraco.com. Picked up the other domains in the master template while I was there.


nul800sebastiaan:
Great, thanks @ed-parry !!

PR #271, opened 2018-06-28
Updated site URL in the home content

ed-parry:
Another tiny update but a valuable one to not confuse users. Ran a quick check to see if there any further mentions but can't see anything else that isn't relative.


abjerner:
Btw - if you're up for the task, there may still be some occurrences here and there:

https://github.com/umbraco/OurUmbraco/search?q=our.umbraco.org&unscoped_q=our.umbraco.org

Not all of them are relevant to fix, but some are. And some occurrences are probably in legacy code that isn't used anymore.

ed-parry:
Thanks @abjerner - added in a few more. Searched through `cshtml` and `cs` files for any that appear in text, or more crucial links to save on the redirect. Introducing a constant could make sense, especially if it's possible/likely that further changes could be made.

I have another question for @nul800sebastiaan - email addresses used in the `from` fields, sometimes robot@umbraco.org - are these save to switch to .com, or is the email server still using .org? Or are these just placeholders anyway?

ed-parry:
@abjerner @nul800sebastiaan Merged in the latest to resolve conflicts in case this one is still of value :)

nul800sebastiaan:
Awesome, thanks @ed-parry ! I'll deploy this soon!

Issue #264, opened 2018-06-21
The Video count does not seem to be working ( or at least needs updating! )


readingdancer:
I have just looked at the video page to view the #CG18 videos, currently they all say they have been viewed zero times... but if you click through to one of the videos, you will find that they have actually been viewed.

https://our.umbraco.org/videos/

At the time of writing this, the following video has been viewed 96 times:

https://vimeo.com/273473715

What I would expect to see is the counter showing a number at least similar to the real count on Vimeo, perhaps with a sub note to say the counter is updated every X hours or days.

abjerner:
@ed-parry I just had a look at what goes wrong, and it is because the play count isn't returned properly by the Vimeo API (there is a `plays` property, but the value is `null`).

I think the play count was there back in February when we launched the new community section. I will look into why the play count isn't returned ;)

abjerner:
Apparently the play count isn't available if the access token only has the `public` scope, but it is when using an access token with the `private` scope. Which may actually be a bug on Vimeo's end (or silly behavior at best) as my own access token with the `private` scope can access the play count of Umbraco's videos.

@nul800sebastiaan would it be possible to switch to an access token with the `private` scope? If you go to https://developer.vimeo.com/apps/ and select the app used for the Our Umbraco site, you can generate a new access token (the feature lives under the *Authentication* tab, and then at the very bottom of the page).

@ed-parry it might still be a good idea to check whether a given video has any plays, as new videos might not for some time. If you're looking to do another PR, you can find the code for it here (there is already a similar check for likes):
https://github.com/umbraco/OurUmbraco/blob/master/OurUmbraco.Site/Views/Partials/Videos/Home.cshtml#L150

ed-parry:
Makes sense, thanks @abjerner. I guess they're hiding some of the details unless the scope is private, but if it doesn't have to be associated to the upload account doesn't make much sense to me, but can't say I understand the Vimeo API.

Thanks for the pointer too - I may take a look, and could be worthwhile updating the timestamp at the same time - the latest video on the page is only 28 seconds long and so just lists as "0 min" which looks a little strange :)

nul800sebastiaan:
I've now switched to a token that uses the `private` scope as well. Is that all I need to do and then refresh the cache or was there more required?

abjerner:
@nul800sebastiaan should be enough to update the cache ... or just wait a bit as the cache should be automatically updated ;)

readingdancer:
[...]

PR #272, opened 2018-06-29
Added seconds to video duration. Checking for valid play count.

ed-parry:
Related to #264, though the fix for that core issue requires the Vimeo access token to be given the `private` scope.

This PR adds in seconds to the video duration, as the latest video on the page is only 29 seconds long and so is listed as "0 min" which especially alongside the "0 plays" felt like a bug. It now lists the time as "mm:ss", matching how Vimeo displays them.

I've also added a check on the play counter, so it only displays if there are any. More of a quick catch until the token can be renewed but keeps it clean when new videos are published, and matches the likes functionality too.


nul800sebastiaan:
Great, works as advertised! Thanks @ed-parry - this is now live.

PR #274, opened 2018-06-29
Add profile link to member profile pictures on map

matthew-darby:
Thought it might be nice to have a link through to Umbraco member profiles on the map.

### DIFF
--- a/OurUmbraco.Site/Views/Partials/Community/Map.cshtml
+++ b/OurUmbraco.Site/Views/Partials/Community/Map.cshtml
@@ -23,7 +23,7 @@
 
 <script id="map-marker-template" type="x-tmpl-mustache">
     <div style="display:flex; align-items:center;">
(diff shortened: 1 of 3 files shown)


abjerner:
@matthew-darby Thanks for the PR 👍 

Profile links are currently not shown for privacy concerns, so this is actually be design. As such, I don't think your changes will be accepted.

But to improved your the chances of your PR getting accepted, one way to work with the privacy concerns could be to include an opt-in checkbox so members actively have to select that whether they want their profile link shown as part of the map. @nul800sebastiaan might have some more thoughts on this.

There are however already some open PRs involving editing member profiles, so it might however be a good idea to wait until they are merged.

nul800sebastiaan:
Yep, Anders is right. We're definitely open to having an opt-in checkbox, so if that can be added to this PR that would be cool!

nul800sebastiaan:
This has been open for a while now and nobody has picked it up. I'll close it for now and put the `idea` label on it so we can find it later as this is still a good idea.

Issue #275, opened 2018-07-04
Single quote is rendered as html code in search result title


bjarnef:
It seems single quotes are rendered as html code in search result title instead of the raw html, e.g. try search for `don't`

![image](https://user-images.githubusercontent.com/2919859/42279623-ae2faf56-7f9e-11e8-8351-43f7b1611820.png)

![image](https://user-images.githubusercontent.com/2919859/42279632-b98485ac-7f9e-11e8-9dd5-1359cd556f67.png)

There is a similar issue in the autosuggestions.

![image](https://user-images.githubusercontent.com/2919859/42279769-271e6786-7f9f-11e8-87e1-9187bdb6d71f.png)

umbrabot:
Hiya @bjarnef,

Just wanted to let you know that we noticed that this issue got a bit stale and might not be relevant any more.

We will close this issue for now but we're happy to open it up again if you think it's still relevant (for example: it's a feature request that's not yet implemented, or it's a bug that's not yet been fixed). 

To open it this issue up again, you can write `@umbrabot still relevant` in  a new comment as the first line. It would be *super* helpful for us if on the next line you could let us know why you think it's still relevant.

For example:

>@umbrabot still relevant
>This bug can still be reproduced in version 8.9.0

This will reopen the issue in the next few hours.

Thanks, from your friendly Umbraco GitHub bot :robot: :slightly_smiling_face:

PR #279, opened 2018-07-13
Issue #275: Updated search to decode HTML strings before displaying

ed-parry:
Fixes issue #275 - I think the logic was wrong here - it should encode the strings on the way up, and decode them on the way down? So I've switched it over when rendering the search results, and removed the escaping step from the AJAX search drop-down as well, but shout if I'm miles off the mark here - assuming the text is encoded/escaped before being stored, so in the AJAX case we shouldn't need to escape again (and removing it allows the necessary characters to come through okay.


ed-parry:
@nul800sebastiaan Hey Sebastiaan - looks like a merge is needed here. Just wanted to check if (after resolving the conflicts) this would be good to go?

nul800sebastiaan:
Hey @ed-parry! I am actually unsure if we should do this, it could lead to people being able to craft some javascript to do nasty things.  🤔

ed-parry:
Morning @nul800sebastiaan. Yeah, I know what you mean. My thinking here is that the inputs are being checked and sanitized before going into the examine indexes, so it _should_ be safe, but I'd be lying if I said I was sure. Can take another look to see if there's a better way, or at least test further with JS examples?

nul800sebastiaan:
Yeah I'm not sure if that is true, we do some sanitizing when people post things to the forum but I have also seen it break some things. I'd hate to break the whole search just because 1 forum post always comes up. 

Although, the auto suggest search (see #275) gets the titles correct so maybe you can double check what we do there?

PR #281, opened 2018-07-19
Add rewrite for deleted courier docs

hemraker:
see https://github.com/umbraco/UmbracoDocs/pull/930 for ref


nul800sebastiaan:
Beter late then never 🙈

PR #282, opened 2018-07-21
Updated headline variable conditional

jonrandahl:
Currently, the headline for both the "Updated Projects" AND "Popular Projects" filters show the same message: "Browse most popular projects"; this checks if the value passed back from `Request["orderBy"]` is actually "popularity" and assigns the corresponding message correctly.

I've commented out the old code to show that the "latest" message would never actually be shown with that logic in place.

Cheers all!


jonrandahl:
Closing as outdated

PR #306, opened 2018-09-11
fix #283

### LINKED ISSUE #283: Update name reply-name
I know @leekelleher is famous, MVP and awesome contributer to Umbraco, but can't I just reply **Lee Kelleher** instead of **Lee KelleherMVP 8xadminc-trib**? 😁🦄 

image

villian:
fixed retrieving of a recipient name when you write a replay on topic for somebody

`.closest('.actions').prevAll('.meta').find('a').first().text();`


nul800sebastiaan:
Thanks @villian !

PR #329, opened 2018-10-24
Consistent styling for member badges

mzajkowski:
I've found inconsistency between styling of badges for the member on Our. It had different styles and states on various places, so I've decided to fix it.

This is how it looks now (before fix, outiside of the thread): https://cl.ly/81c4f74b0943
In the thread: https://cl.ly/9b932c75f734

---

And after small fixes..

Profile: https://cl.ly/c496799f2003
Thread: (stays same)

---

I was trying to fetch the thread styling - dunno if it's the desired / consistent one to keep. We can always get back to colored version of it as well, but I was really annoyed by the underline kept there no matter of hovering over or not.

Feedback more than welcome!

### DIFF
--- a/OurUmbraco.Client/src/scss/elements/_member.scss
+++ b/OurUmbraco.Client/src/scss/elements/_member.scss
@@ -41,9 +41,7 @@
 		}
 
 		a {
-			&:hover {
-				text-decoration: none;
-			}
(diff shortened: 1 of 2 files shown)


nul800sebastiaan:
Cool, thanks @mzajkowski !

PR #334, opened 2018-10-27
Update mapR.js

xterionX:


abjerner:
Hi @xterionX  and thanks for creating this this PR. It's always great to see new users contribute to the Umbraco repositories 👍 

I'm however a bit unsure why you made this change. Can you elaborate a bit on this?

The `let` keyword was introduced so you could declare variables with a scope at the block level - instead of `var` which is either global or at the function level. From what I can tell, all the variables you changed are already either at the global level or function level, so I don't think switching to `let` would make any difference.

While there may not be many users left using Internet Explorer, it doesn't have full support for `let` - which may be another reason to stick with the current `var`.

nul800sebastiaan:
I'm sorry @xterionX - @abjerner is right, there is no need to change this file with the proposed changes. 

I'll close this and also #333 

Thanks!

PR #338, opened 2018-10-30
Would .info and .notice make sense as well?

abjerner:
I have a specific case for the documentation where I'm describing that a feature isn't available in earlier versions of Umbraco. I don't really feel *Tip* is the best title here, so my perhaps we should add `.info` and `.notice` as well?

image

I'm not that familiar with the `:::` syntax. Would it be possible to specify the title via Markdown syntax?

### DIFF
--- a/OurUmbraco.Client/src/scss/sections/_docs.scss
+++ b/OurUmbraco.Client/src/scss/sections/_docs.scss
@@ -116,6 +116,13 @@
         @include notice(rgba(40,51,204,.07), rgba(40,51,204,.70), "\e148  Tip:");
     }
 
+    .info {
+        @include notice(rgba(40,51,204,.07), rgba(40,51,204,.70), "\e148  Info:");
+    }
+     .notice {
+        @include notice(rgba(40,51,204,.07), rgba(40,51,204,.70), "\e148  Notice:");
+    }
+
     .checklist {
         ul {
             list-style: none;


dampee:
Hi Anders,

Why would would "memo" not work for your case?

The `:::` syntax just adds a class in the css.  Nothing more.  The title is currently added by css.  It's an implementation of @jmayntzhusen.  I don't think the MarkDig library supports a "title" to be added.
We could suggest to change the implementation to a "colour" and add the title in markdown.
I haven't tried to write a new plugin for markDig, but that could work too.  :-)

Adding a "notice" and "information" makes sense in my eyes.

In this particular case, if it is for a newly introduced methods in a particular version, it is possibly more important to create a new file for the version where it was introduced.  If you need any help with that let me know. (or point out to the PR you have prepared.

nul800sebastiaan:
`info` seems fine to me, `notice` always reads like "This is a public service announcement, we WILL be testing the fire alarm at 12.00. This is a drill.". `note` seems better?

jmayntzhusen:
In the end it's just a few lines of CSS, so if anything is made more clear by adding a few extra fields it doesn't really hurt? 🙂

Are you sure you want them to have the same color and icon?

nul800sebastiaan:
Wrong question: what is the benefit? :)

abjerner:
Although it will be visually the same in terms of color, I think seeing a box with **Notice** will give the box a higher significance for the reader than just **Tip**.

It's not a big thing, but like I wrote in the original description, I didn't think **Tip** was the best title. **Note** could also work instead - or perhaps replace **Tip** as I see **Note** have a broader use.

jmayntzhusen:
Note already exists alongside tip @abjerner 🙂

The three current notices are here: https://our.umbraco.com/documentation/Contribute/Markdown-Conventions/#adding-notes-warnings-tips

abjerner:
@abjerner I hadn't noticed `note`. Has that been added since I created this PR?

`note` is very similar to `notice`, so I think that is sufficient.

jmayntzhusen:
Has been there the whole time but used to be green 🙂

PR #339, opened 2018-10-30
Corrected class name...

dbeezt:
...although if it's just an empty div it might warrant being deleted?


nul800sebastiaan:
Thanks @dbeezt - have you tested this change to make sure this doesn't break anything? Can you please provide steps to test?

Issue #324, opened 2018-10-18
Umbraco 7 tag doesn't link to any page


bjarnef:
For this forum thread it has two categories/tags: "Using Umbraco And Getting Started" and "Umbraco 7".
https://our.umbraco.com/forum/using-umbraco-and-getting-started/89962-image-preprocessing-on-upload-to-media

The first one links to https://our.umbraco.com/forum/using-umbraco-and-getting-started/ but the second one doesn't link to any page.

![image](https://user-images.githubusercontent.com/2919859/47140552-d725d480-d2be-11e8-8e6f-691a8dc9fc6c.png)

kumaran-14:
If the link is provided i could make a pr.

sarnava1:
@bjarnef @nul800sebastiaan please assign this issue to me. I am working on this.

sarnava1:
@bjarnef @nul800sebastiaan please can you give me the link where the anchor tag should refer to?

sarnava1:
@bjarnef @nul800sebastiaan shall I remove the href tag?

umbrabot:
Hiya @bjarnef,

Just wanted to let you know that we noticed that this issue got a bit stale and might not be relevant any more.

We will close this issue for now but we're happy to open it up again if you think it's still relevant (for example: it's a feature request that's not yet implemented, or it's a bug that's not yet been fixed). 

To open it this issue up again, you can write `@umbrabot still relevant` in  a new comment as the first line. It would be *super* helpful for us if on the next line you could let us know why you think it's still relevant.

For example:

>@umbrabot still relevant
>This bug can still be reproduced in version 8.9.0

This will reopen the issue in the next few hours.

Thanks, from your friendly Umbraco GitHub bot :robot: :slightly_smiling_face:

PR #340, opened 2018-10-30
Include "Version" attribute in TopicController

dbeezt:
#324 - Hopefully a step closer to finding a solution. It seems that all of the other model attributes (located at https://github.com/umbraco/OurUmbraco/blob/master/OurUmbraco/Forum/Models/TopicSaveModel.cs) are called in the Post method while Version has been left out for some reason. Apologies if this isn't helpful but I'd be keen to understand why it's been left out of the controller despite being included in the model?

### DIFF
--- a/OurUmbraco/Forum/Api/TopicController.cs
+++ b/OurUmbraco/Forum/Api/TopicController.cs
@@ -16,6 +16,7 @@ namespace OurUmbraco.Forum.Api
             t.MemberId = Members.GetCurrentMemberId();
             t.Created = DateTime.Now;
             t.ParentId = model.Forum;
+            t.Version = model.Version;
             TopicService.Save(t);
         }
 


nul800sebastiaan:
@dbeezt We currently don't have any search results for umbraco versions so we can't really like to anything. So I'd prefer to just remove the link for now. Adding an option to get all topics marked with a version is going to require updated to the search.

Mostly, the version number is just there to provide an indication of what Umbraco version we're talking about and at the moment is not very relevant. 99% of the topics are about v7.

dbeezt:
Hey, so I've edited the version tag's link:hover and href via [inspect element](https://user-images.githubusercontent.com/10518422/47810968-7ecde880-dd3c-11e8-9460-2aac5541de91.png) but I can't for the life of me find the actual .css file in the OurUmbraco repo - don't suppose you'd know the location?

nul800sebastiaan:
Hi @dbeezt and very sorry for the late reply 🙈 

The css is generated from some scss files in the `~/OurUmbraco.Client/src/scss` folder. Will you have a chance to update that one? It should probably just be changed from an `<a/>` tag to a `<span/>`.

In order to build the CSS changes you'd need to run a `gulp build`.

nul800sebastiaan:
I'll close this one for now, @dbeezt - feel free to open a new one if you still want to work on this, thanks! 👍

PR #346, opened 2018-11-08
Increasing the db connection timeout in template web.config

LottePitcher:
When first getting the Our website to run locally the database reported timeout issues as it has a lot of upgrading/migrations to do.  The default connection timeout is 15 seconds.  I've set it to be 5 minutes in the template web.config.


nul800sebastiaan:
Because patience is a virtue!

PR #349, opened 2018-11-14
Improved colors for XML syntax highlighting

abjerner:
Seems I didn't test the colors for XML syntax highlighting for my other PR, as the colors are better for light themes. So with this PR, I'm turning:

image

into:

image

The latter is based on the default theme from VS Code, which has a dark background similar to the Our forum.


nul800sebastiaan:
Sweeeeet!

PR #352, opened 2018-11-25
Remove Yahoo Pipes from master.cshtml

dampee:
Yahoo pipes does not exist anymore, let's save bandwith :-)


nul800sebastiaan:
Haha, yes, let's! Thanks! 👍

PR #353, opened 2018-11-25
Remove the trailing spaces from yaml markers

dampee:
As this is hard to find out problems in the yaml as github does not shows this.

Just spend an hour on trying to find out what was happening with YAML not being interperted.  It was a space after the "---" marker.


nul800sebastiaan:
👍

PR #360, opened 2018-12-03
Updated the syntax highlighter

abjerner:
As the JSON examples used throughout the documentation doesn't follow the JSON specification 100%, I realized I had to adjust my syntax highlighter package a bit (see some discussion [here](https://github.com/umbraco/UmbracoDocs/pull/1250)).

Where the old version of my package used the old ColorCode from Microsoft (which had some bugs on it's own), I decided to write my own JSON tokenizer/parser, which then gives a number of benefits - eg. comments and strings that are not properly enclosed with double quotes.

As GitHub doesn't like invalid JSON either, @jmayntzhusen suggested we could use `json5` (human friendly JSON, that is) instead of `json`, I've also updated my Markdig extension and the code for Our to allow setting `json5` as an alias of JSON.

### DIFF
--- a/OurUmbraco.Client/src/scss/elements/_markdown-syntax.scss
+++ b/OurUmbraco.Client/src/scss/elements/_markdown-syntax.scss
@@ -534,6 +534,7 @@
         .keyword { color: #a71d5d; }
         .constant { color: #b5cea8; }
         .string { color: #ce9178; }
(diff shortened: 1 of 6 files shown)


nul800sebastiaan:
Sweet! Thanks @abjerner !

PR #361, opened 2018-12-04
Added SEO meta data to indexing with fallback

dampee:
* new YAML tags on documentation markdown files
* make it possible to control page Title and Meta Description on documentation yaml

PR #364, opened 2018-12-11
Make search titles more friendly in the documentation

jmayntzhusen:
This will make a page like: 
https://our.umbraco.com/documentation/getting-started/setup/install/system-requirements

That has a title like this: 
system-requirements - install - setup - getting-started - Documentation - our.umbraco.com

Turn into something like this:
System-Requirements, Getting-Started - Our Umbraco

So search results will look a bit nicer than this:
image

Long term we want to change how this works with https://github.com/umbraco/OurUmbraco/pull/361

But for now this can make it better in the meantime

### DIFF
--- a/OurUmbraco.Site/Views/Master.cshtml
+++ b/OurUmbraco.Site/Views/Master.cshtml
@@ -75,7 +75,7 @@
                 {
                     var path = title.Split(new string[] { " - " }, StringSplitOptions.None);
 
-                    System.Globalization.TextInfo myTI = new System.Globalization.CultureInfo("en-US", false).TextInfo;
+                    var textInfo = new System.Globalization.CultureInfo("en-US", false).TextInfo;
 
                     var firstUrlPath = myTI.ToTitleCase(path[0]);
                     var lastUrlPath = myTI.ToTitleCase(path[path.Length - 1]);


OurUmbraco.Site/Views/Master.cshtml
```diff
@@ -71,11 +71,24 @@
 
             if (string.IsNullOrWhiteSpace(title) == false)
             {
-                <text>@title - </text>
+                if (Model.Content.Name.ToString().ToLower() == "documentation")
+                {
+                    var path = title.Split(new string[] { " - " }, StringSplitOptions.None);
+
+                    System.Globalization.TextInfo myTI = new System.Globalization.CultureInfo("en-US", false).TextInfo;
```
nul800sebastiaan: Just use `var` and make sure to give your variable a meaningful name, I would suggest `textInfo`.

OurUmbraco.Site/Views/Master.cshtml
```diff
@@ -71,11 +71,24 @@
...
+
+                    var firstUrlPath = myTI.ToTitleCase(path[0]);
+                    var lastUrlPath = myTI.ToTitleCase(path[path.Length - 1]);
```
nul800sebastiaan: Maybe also replace `-` with an empty string?

jmayntzhusen: Not sure what you mean? 🙂

nul800sebastiaan: Now the title will be `System-Requirements, Getting-Started`, why not `System requirements, Getting started`?

Issue #365, opened 2018-12-18
Contributors list on homepage is only collecting first 100 contributors for each repo


stevetemple:
The endpoint being used is returning a maximum of 100 results and doesn't support paging:
https://api.github.com/repos/umbraco/Umbraco-CMS/stats/contributors

I'm going to try changing the call used to
/repos/:owner/:repo/contributors

Or alternatively try the v4 GraphQL API

nul800sebastiaan:
@stevetemple Yep, we already have all of the data downloaded as json files for the release notes. Please don't start with GraphQL yet, we have nobody to review it properly due to missing knowledge! :-)

Unfortunately, you won't be able to download all the data right now but I'll see if I can post a zip.

nul800sebastiaan:
This needs to be unzipped in: `OurUmbraco.Site\App_Data\TEMP\`

https://umbracoreleases.blob.core.windows.net/ourumbraco/GitHub.zip

It should look something like this:

![image](https://user-images.githubusercontent.com/304656/50156792-f695a980-02cf-11e9-8f8e-98a1492df00c.png)

stevetemple:
> Please don't start with GraphQL yet, we have nobody to review it properly due to missing knowledge! :-)

Ahh that seemed like the fun part!

stevetemple:
Oh, we really should try the GraphQL API it's actually umbazing, just been having a play as it's something I want to start doing anyway. Head over to https://developer.github.com/v4/explorer/ and stick in a query like 

`query {
  repository(owner: "Umbraco", name:"Umbraco-CMS")
  {
  	pullRequests(first:100, states: MERGED) {
      edges
      {
        node{
          number
          url
          author {
          	login
        	}
      	}
        cursor
      }
  	}  
  }
}`

nul800sebastiaan:
Looks good, no exclude all the HQ members and the people who are in the PR teams? 

Lists:
-  https://github.com/umbraco/OurUmbraco/blob/master/OurUmbraco.Site/config/TeamUmbraco.json
- https://github.com/umbraco/OurUmbraco/blob/master/OurUmbraco.Site/config/githubhq.txt

And then.. you know, building this query from C# somehow 😁

readingdancer:
@abjerner Hi, I noticed that the contributor list on Our doesn't update when you get V8 pull requests merged in, as you've probably seen I have submitted a few recently and was surprised the count didn't increase on the Our home page, now I know why :)

[...]

PR #366, opened 2018-12-18
Update GitHubService to use async/await for RestSharp methods

stevetemple:
I was investigating why some new contributors were not showing on the homepage. 

I tried this to eliminate race conditions. Seems like an improvement even if it doesn't fix the underlying issue.


nul800sebastiaan:
@stevetemple Before you spend a lot more time on this, note that the homepage list needs to be replaced completely. It only shows new contributors to certain branches and it doesn't show when we do a "squash and merge" when merging pull requests.

While these changes might be fine (haven't looked), we need to use the other github integration that we use for PRs to make the release note (for example).

stevetemple:
Sure I was actually investigating something which I've captured here https://github.com/umbraco/OurUmbraco/issues/365

Issue #320, opened 2018-10-12
Crop size of image used as package icon


bjarnef:
I think we should crop the size of the image, which are use as package icon, which appear in the results of packages.

E.g. try searching for "h-tag":

![image](https://user-images.githubusercontent.com/2919859/46856926-01cbe500-ce08-11e8-90f0-19dfe15c86ab.png)

E.g. this default image (icon) for this package, breaks the layout a bit.
https://our.umbraco.com/packages/backoffice-extensions/tag-manager/

nul800sebastiaan:
Hehe, yeah that looks less than ideal! 

I've marked this as "Up for grabs" so that you or someone else coming along could create a pull request for it.

Matthew-Wise:
This can be closed as the PR is merged and live :)

PR #373, opened 2019-01-09
Small fix for the icons shown in the package search results.

readingdancer:
This is a fix for issue #320 

Rather than cropping as the ticket suggested, I have just set the max height and width to be 64px square as this was already the max width. The problem with setting a crop is the package owner would have no control over how their image is cropped. This way, we can inform package owners of the exact dimensions and it's up to them to upload an image that works.

### DIFF
--- a/OurUmbraco.Client/src/scss/elements/_packages.scss
+++ b/OurUmbraco.Client/src/scss/elements/_packages.scss
@@ -86,8 +86,8 @@ body.packages {
 		}
 
 		img {
-			max-width: 64px;
-			max-height: 64px;
+            max-width: 100%;
+            height: auto;
 			display: block;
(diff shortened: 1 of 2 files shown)


nul800sebastiaan:
Hmm, not sure this is the best solution, it would turn into some weird resizing. Maybe it would be better to do the same as in the non-search overview and just scale it as square with different sizes for different resolutions?

image

readingdancer:
Hi @nul800sebastiaan 

I've changed it as per your suggestion :)

I also made them both 64x64 instead of the 50x50 which was being used as they were being stretched to that size in the browser.

Cheers,

Chris

nul800sebastiaan:
Great, thanks for the update! I'll deploy this soon.

PR #375, opened 2019-01-13
Fixed issue with GitHub usernames

abjerner:
Reported here: https://twitter.com/waltza86/status/1084047844464570368

The issue was caused by two bugs:

**IIS rewrite rule**
The rewrite rule would match usernames against `(\w+)`. `\w` is short for `a-zA-Z0-9_`, which doesn't include hyphens. The regex also didn't wrap to the end of the line, which was why it still matched `Matthew` even though the URL contained `Matthew-Wise`. Both should be fixed now.

**Search in Examine**
The search in Examine may reveal partial matches. So when searching for `github:Matthew`, Examine would return more than one result, and we would just pick the first and display the profile for that member.

With the PR, we're doing an extra check in C#/Razor to verify show the profile for the correct member.

### DIFF
--- a/OurUmbraco.Site/Views/Partials/Members/PublicProfile.cshtml
+++ b/OurUmbraco.Site/Views/Partials/Members/PublicProfile.cshtml
@@ -32,11 +32,17 @@
         // Make the search in examine
         ISearchResults results = searcher.Search(criteria);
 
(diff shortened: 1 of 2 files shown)


nul800sebastiaan:
Thanks @abjerner ! I'll deploy later today! 👍

Issue #384, opened 2019-01-28
Update Gulp and all the client side build dependencies.


readingdancer:
Gulp was using version 3.x and the latest version is currently 4.0 there were a lot of security vulnerabilities showing when you try and build the client side files.

Also the build script includes imagemin but is not using it, for some reason it is only compressing SVG images... twice!

Finally it is possible to run Gulp processes in Parallel, so assuming you have a multi-core machine you should see a performance increase in the build process.

nul800sebastiaan:
Fixed in https://github.com/umbraco/OurUmbraco/pull/385

readingdancer:
Closing the issue as it's now been merged in.

PR #385, opened 2019-01-28
Updating Gulp and the client side build process.

readingdancer:
This Pull request fixes the issues raised in issue #384 

- Updated Gulp to version 4.0
- Added a DEV gulp task that builds everything except the images  ( FYI, on my laptop this now takes 1.2s )
- Fixed the Imagemin section so it's now trying to compress all the image files, previously it was trying to only compress SVG's and it was doing it twice!
- Updated all the other dependencies so there are now no Vulnerability errors
- Made the DEV & BUILD tasks run their child tasks in Parallel so it should run a little faster on a multi-core machine.
- Added a couple of IF statements that prevent errors being shown when the background tasks are trying to fetch the YouTube and Vimeo Videos data via their respective API calls, these errors just show up in development if you don't have API credentials setup in the Web.config file


nul800sebastiaan:
Thanks, this is just to acknowledge I've seen it and to let you know it's going to take a few weeks to get back to you on this one. I need to make sure this generates that same output and.. most importantly, the build server can use it (which I expect to be "interesting").

readingdancer:
As they say "it works on my machine" ;-)

I know what you mean about your build server, no doubt it might need NodeJS and NPM to be updated to the latest versions.

FYI, I am currently using:
- Node v10.15.0
- NPM v6.4.1

Other than that, it does all seem to be working find locally, so I am pretty sure once you have it building it will be fine. I am sure as the libraries were quite out of date that there will probably be a few small improvements in the output, so I wouldn't expect the output to be identical.

I have pushed up a couple more changes to this branch, removing a couple of old libraries and replacing them with their modern counterparts. Also removed two libraries that we were not using at all.

Again I've tested the site locally and all seems fine on my end :)

nul800sebastiaan:
Thanks again @readingdancer - this seems fine and works on my machine too.. Let's keep our 🤞 for the build server! :)

nul800sebastiaan:
It worked, whoop!! 🎉

PR #390, opened 2019-02-11
Changes for v8 documentation

jmayntzhusen:
These changes will index all non-versioned articles as both v7 and 8. Set the default search bar to v8.

This will be accompanied by a Docs pull request renaming all index-v8.md files to index.md, and a few other small things. Will reference it here once it is done.

**This is not to be merged before the v8 launch**


jmayntzhusen:
Closing as #392 and #393 are better versions of this 🙂

PR #391, opened 2019-02-11
Base contributors on merged PRs not commits

stevetemple:
I now fetch pages of PR data using GitHubs GraphQL API to get all PRs. This is run as a nightly Hangfire Schedule, because of throttling at GitHub I'm doing each page/repo request as new task so it's got built in retrying and a small delay. This took about 5-10 minutes to fetch the initial load of all PRs.

I'm storing these in a new DB table as it seemed like the most sensible place to keep the history. 

Currently the read code is hitting the DB each time, we might want to either outputcache or application/http/memory cache it.

It's showing the number of merged PRs for each user and then the sum of additions and deletion within those PRs

Fix for : https://github.com/umbraco/OurUmbraco/issues/365


stevetemple:
I'm still pursuing this, but taking a different approach, another PR will be incoming soon

nul800sebastiaan:
That's cool, thanks @stevetemple - can we find a solution where we don't need to use database tables, that will make it a whole lot easier to deploy 🙈

Issue #382, opened 2019-01-25
Versions box needs z-index: 1


BatJan:
Going through a bunch of documentation I stumbled upon this issue, which I by mistake originally posted at the issue tracker for the documentation here https://github.com/umbraco/UmbracoDocs/issues/1461 - So the below is just copy/pasted directly and edited a bit.

**What type of issue is it?**
Other: Minor CSS bug

**What article/section is this about?**
[Dropdown](https://our.umbraco.com/documentation/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Dropdown)

**Describe the issue:**
The versions box needs position: relative and z-index: 1; since there otherwise is a risk that the headlines will lay on top of it. Try clicking the last "Current" link in the versions box in the above link - Currently not possible (On desktop at least) 😃

nul800sebastiaan:
Thanks @BatJan ! 

I've marked this as "Up for grabs" so that you or someone else coming along could create a pull request for it.

nul800sebastiaan:
Jesper applied it and this is deploying now. 👍

PR #392, opened 2019-02-20
cross search version 7 and 8 on umbraco docs

dampee:
Because all files have now a versionFrom (when https://github.com/umbraco/UmbracoDocs/pull/1574 is merged) we can turn the logic around to make sure that the file can be found a cross multiple versions.

To test: 
Search for Tours (v7) will return content.  Currently "Tours" with v8 won't return content.
If this PR (and the docs PR) is merged, the v8 will return the same content as v7 as there is no difference in the documentation for the "tours" section.  If you search for v6, there should be no results.

### DIFF
--- a/OurUmbraco.Client/src/scss/sections/_docs.scss
+++ b/OurUmbraco.Client/src/scss/sections/_docs.scss
@@ -86,6 +86,8 @@
     flex-wrap: wrap;
     text-decoration: none;
     background-color: white;
+    z-index: 1;
+    position: relative;
 }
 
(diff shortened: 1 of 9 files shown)


dampee:
If this is merged in, you can close #390

jmayntzhusen:
Combined this one and #393 except for the web.config change which will be set by @nul800sebastiaan .

Also added support for the versionRemoved feature in the Docs and fixed https://github.com/umbraco/OurUmbraco/issues/382

Should be ready for testing now 🙂

PR #393, opened 2019-02-20
Prepare major version switch for docs

dampee:
What needs to be updated when v8 will be launched to switch the docs to a new major


dampee:
If this PR and #392 are merged, you can close #390

jmayntzhusen:
These changes have been included in #392, will close this now 🙂

PR #400, opened 2019-02-27
CSS updates fixing issues on mobile and member profile

Matthew-Wise:
Fixed issue with mobile navigation been white on white until the sticky bar appears. 

Fixed navigation levels header tags to be white on the mobile blue backgrounds.

Changed the green solved to the blue on the member profile.


nul800sebastiaan:
Very lovely, thanks very much! And thanks for the reminder on umbraCoffee! 👍

PR #406, opened 2019-03-14
Add Moriyama blog

darrenferguson:
Add Moriyama blog to community blogs

### DIFF
--- a/OurUmbraco.Site/config/CommunityBlogs.json
+++ b/OurUmbraco.Site/config/CommunityBlogs.json
@@ -252,7 +252,7 @@
       "logo": "https://pstokes.co.uk/media/1017/img_2254-3.jpg",
       "memberId": 146507
     },
-     {
+    {
       "id": "97488df9-0f87-4748-b8cc-b68ca48b2030",
       "checkTitles": false,
       "title": "Owain Codes",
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
@darrenferguson Thanks! Unfortunately, we currently can't parse "Tue, 12 Mar 2019 12:17:36 EST" - instead of `EST` we're expecting a UTC offset (in your case `+00:00`). Any chance you can update your RSS feed to use that instead?

abjerner:
@nul800sebastiaan I may have a fix for that. IIRC the `EST` is technically valid, but .NET fails to parse it.

abjerner:
Our Umbraco already has a dependency on my `Skybrud.Essentials`, so using this method for parsing the date should work:

https://github.com/skybrud/Skybrud.Essentials/blob/v1.1.5/src/Skybrud.Essentials/Time/TimeUtils.Rfc822.cs#L54

I will look into creating a PR for the current RSS parsing logic 😉

darrenferguson:
@nul800sebastiaan fixed!

nul800sebastiaan:
@abjerner Ha! Ironically, we do use your lib.. but then we parse it as a datetime offset.. ;)

image

nul800sebastiaan:
@darrenferguson cool, works now! 👍

abjerner:
@nul800sebastiaan sure that is the code that fail? `Rfc822ToDateTimeOffset` works fine with `Tue, 12 Mar 2019 12:17:36 EST` when I call it directly.

On the other hand, the `ParseRfc822DateTime` method in the class below doesn't use my helper method:

https://github.com/umbraco/OurUmbraco/blob/master/OurUmbraco/Community/BlogPosts/BlogRssUtils.cs

nul800sebastiaan:
@abjerner `DateTimeOffset.TryParse` is what fails there since the `publishDate` still contains `EST`.

abjerner:
@nul800sebastiaan a missing return state could have something to do with it 😮 

PR is here: https://github.com/umbraco/OurUmbraco/pull/407

PR #407, opened 2019-03-14
Fixed RSS date parsing

abjerner:
As reported in #406, the date parsing in RSS feeds didn't quite work as expected.

My `TimeUtils.Rfc822ToDateTimeOffset` method used for the parsing actually does what it should, but the returned value was never used as it was immediately overwritten by `DateTimeOffset.TryParse`.


nul800sebastiaan:
Duh! Nice one @abjerner - Thanks very much! 👍

PR #408, opened 2019-03-15
Make removing the v8 warning note possible

jmayntzhusen:
# Why?

It is currently not possible for the Docs Curators to remove the note at the top warning people that the article hasn't been tested for v8. We had a discussion at our last meeting and would like to set a YAML tag on all the sections where it makes sense, then those can show the note - it would work as a sort of reminder of what is still in need of an update / check.

# How?

Have added functionality where a YAML tag can be set and then it will only show the note if that is not set. To enable the note we would simply set `needsV8Update: "true"`. We will shortly be rolling out a Docs PR that will add this tag to all articles that currently has the note 🙂 Which is why this is currently a draft PR.

### DIFF
--- a/OurUmbraco.Site/Views/Partials/Documentation/DisplayMarkdown.cshtml
+++ b/OurUmbraco.Site/Views/Partials/Documentation/DisplayMarkdown.cshtml
@@ -157,10 +157,7 @@
 
 @if (currentUmbracoVersion == 8
      && currentVersionIsRemovedForv8 == false 
-     && currentVersion != null
-     && currentVersion.Url.EndsWith("-v7") == false
-     && currentVersion.VersionFrom.Major == 7 
-     && currentVersion.VersionTo.Major == 0)
+     && currentVersion.NeedsV8Update == "true")
 {
     <div class="note">
         <p><strong>This article has not yet been verified against Umbraco 8.</strong></p>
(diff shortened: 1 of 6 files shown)


nul800sebastiaan:
What would it do?

Either you've verified that a piece of documentation works for v8 or you have not verified it:
- If verified: change the version range (either to include v8 or to stop at v7)
- If not verified: the warning shows

What am I missing?

jmayntzhusen:
The issue is that when you verify it works for v8, you add versionTo: 8.0.0, then 8.1.0 comes out and you need to go through a bunch of files and update it to 8.1.0, etc. 

With this change we can leave it open ended so versionFrom: 7.0.0 fx would mean from that to current. It would also give us an easier overview over what needs attention as we can clone the docs down and search for files with the YAML tag at the top. 

Once we are down to 0 of them we can remove the YAML tag from the index, etc again 🙂

nul800sebastiaan:
@jmayntzhusen Then set it to 8.99.99

But fine, we'll put a hack upon the hack but remember you'll need to  manually undo all of this weird tagging at some point again.

jmayntzhusen:
Issue with setting it to 8.99.99 is also when we then add a versioned article later on this number will show up on the pick version box 😬

Yes we will remove them all once we have none 🙂

jmayntzhusen:
This is now ready to be merged as https://github.com/umbraco/UmbracoDocs/pull/1623 is already set up for it 🙂

Issue #404, opened 2019-03-14
Mismatch between label and input


kjsguld:
Page in question: [https://our.umbraco.com/member/forgot-password/](https://our.umbraco.com/member/forgot-password/)

The label for the field asks for "Username", but validation asks for "Email address"

![image](https://user-images.githubusercontent.com/19648887/54339188-5fbd5c80-4634-11e9-8d0c-988208445f7c.png)

Reproduction steps: 
enter "username" in field

Fix: 
Change label to say "Email address"

nul800sebastiaan:
Relates to https://github.com/umbraco/OurUmbraco/issues/398

nul800sebastiaan:
Fixed in https://github.com/umbraco/OurUmbraco/pull/412

PR #411, opened 2019-03-21
Changed "UserName" to read "Email address". Also fixed a typo "cless" to "class"

poornimanayar:
This PR fixes issue #404
There was also a typo "cless" by the input label on forgot password and reset password forms which I have fixed. Also in addition to the forgot password page I have renamed "Username" to "Email address" as the field label on login page to ensure consistency


nul800sebastiaan:
Fixed in https://github.com/umbraco/OurUmbraco/pull/412

Issue #401, opened 2019-02-28
Packages page under profile is broken


bjarnef:
When navigating to your Our profile the left navigation has a menu item "Packages" which link to `/member/profile/packages/`.

![image](https://user-images.githubusercontent.com/2919859/53549519-a8552000-3b34-11e9-8509-0302add3059c.png)

However it seems this page in broken since it returns 302 Found and shows this page.

![image](https://user-images.githubusercontent.com/2919859/53549634-f9651400-3b34-11e9-90fe-97797019a8e8.png)

Btw. it would be great with a nice 404 and general error page (maybe with a unicorn or something else 🦄).

nul800sebastiaan:
Thanks, related to the Cloud outage at the moment.

nul800sebastiaan:
Ah nope, sorry different issue. No idea why yours doesn't work, mine does. 🤔

nul800sebastiaan:
Some kind of null reference error on `GetListingsForContributor` - will get back to this later.

nul800sebastiaan:
Fixed in https://github.com/umbraco/OurUmbraco/pull/412

nul800sebastiaan:
This has been deployed now, did it work @bjarnef ?

bjarnef:
@nul800sebastiaan well, I only see the "profile" link in left navigation, so I can't check what the previous "packages" link does?

![image](https://user-images.githubusercontent.com/2919859/56612412-a5e2e400-6614-11e9-9342-f8ac5c5f8e76.png)

nul800sebastiaan:
@bjarnef That.. is very weird! It should just list the children of a certain node. What happens when you go to https://our.umbraco.com/member/profile/packages/ ?

bjarnef:
@nul800sebastiaan on packages page it shows this.

![image](https://user-images.githubusercontent.com/2919859/56655044-fea49200-6691-11e9-9949-5afb52e42872.png)

On this page https://our.umbraco.com/member/profile/

![image](https://user-images.githubusercontent.com/2919859/56655186-62c75600-6692-11e9-80c0-f07d128b9440.png)

It was just the public profile page (available as link via right offcanvas menu)
https://our.umbraco.com/members/bjarnef/

Maybe the public profile should also have a page "packages" so community users can see which packages a member has created and contributes to?

nul800sebastiaan:
Okay, so it works again, good!

Packages created is there, packages contributed to not yet, think there's a feature request for that already 

[...]

PR #412, opened 2019-03-21
401 : Null check contributed packages pages

poornimanayar:
I think this could be a potential fix to issue #401. I say potential as I am unable to test this locally. But going through the conversation @bjarnef on the issue and the fact that one person experiences it while other doesn't makes me think so.


nul800sebastiaan:
Awesome! I think that will do the trick, you unfortunately added the commits from #411 to this one as well, but I'll just merge this and close the other one, then both are fixed 👍

poornimanayar:
Not again(facepalm)

PR #413, opened 2019-03-22
Revert color change to the note for documentation

jmayntzhusen:
As part of the new Umbraco identity on Our the green note was changed to blue. We already have a blue tip though, I think in this instance it should stay the old color to differentiate 🙂


nul800sebastiaan:
👍

PR #415, opened 2019-03-26
Add v7.14 option to version.config

callumbwhyte:
Umbraco 7.14 was released a couple of weeks ago but currently doesn't show up on the list of available versions for packages... I have updated it here!


nul800sebastiaan:
Whoops, my bad, thanks for that @callumbwhyte  ! 👍

Issue #363, opened 2018-12-05
Breadcrumb for the documentation is too dependent on the URL


abjerner:
The breadcrumb for the documentation is based on the URL of each page, and therefore also reflects the casing of the URL. This may not be super important, but it differs from rest of Our, which uses lower case URLs.

As a versioned page is just another file, the name is also part of the URL - eg. `index-v8`. With this being part of the URL, it also shows up in the breadcrumb.

![image](https://user-images.githubusercontent.com/3634580/49529239-e20be700-f8b5-11e8-90b6-03da6dbb49ab.png)

A possible fix could be to build a sitemap of all pages - possibly also extracting the actual title from either the YAML or `<h1>` of each Markdown file. Any thoughts?

nul800sebastiaan:
Thanks @abjerner - I think the Docs team is working on something like this already. Any feedback @jmayntzhusen ?

jmayntzhusen:
We have been working on something similar with seo titles / descriptions, but we dont have anything for breadcrumbs for now. I agree that it could definitely be improved though, and that any "-v" filenames should be changed or removed from the breadcrumb to something better 🙂

nul800sebastiaan:
Should the breadcrumb (in this case) just be "Content Apps (v8)"?

nul800sebastiaan:
Fixed in https://github.com/umbraco/OurUmbraco/pull/416

PR #416, opened 2019-03-28
Remove breamdcrumb for versioned files

jmayntzhusen:
Currently the index filename is included in the breadcrumbs for versioned files, this PR will remove that.

Current state:
Capture

After this fix:
Capture


abjerner:
Awesome 👍 

And just to link your PR with my original issue, it was reported here: https://github.com/umbraco/OurUmbraco/issues/363

PR #417, opened 2019-04-03
Fix broken link after renaming branches in the CMS

jmayntzhusen:
The contribute navigation button didn't work as it has the branch name in the link, and the branches were recently restructured with different names.

Claus has fixed it directly on the live site, but this will need to be merged in for the next deployment so we don't overwrite his change with the error again 🙂


nul800sebastiaan:
👍

PR #432, opened 2019-05-19
enclose most of web.config sections to set inheritInChildApplications…

LottePitcher:
… to false

Needed so that Our can have a child application running in its own application pool (for docfx version of documentation)


emmaburstow:
Hi @LottePitcher 

Thanks so much for your work here. One of us will take a look and let you know if we need anything further from you.

Emma

nul800sebastiaan:
Thanks! 👍

PR #434, opened 2019-05-20
Label comment automation

emmaburstow:


OurUmbraco/Community/GitHub/GitHubService.cs
```diff
@@ -918,6 +919,47 @@ public void UpdateIssues(PerformContext context, Community.Models.Repository rep
...
+                addComment.CommentBody = addComment.CommentBody.Replace("{{issueowner}}", "@" + recentIssue.User.Login);
+                var commentsEndPoint = string.Concat(resource, recentIssue.Number, "/comments");
+                var request = new RestRequest(commentsEndPoint, Method.POST);
+                request.AddHeader("Authorization", "Bearer 9fb688e28f551ab7278b478145648d09cd6ab662");
+                request.AddHeader("Content-Type", "application/json");
+                request.AddParameter("undefined", JsonConvert.SerializeObject(addComment), ParameterType.RequestBody);
+                client.UserAgent = UserAgent;
+                var result = client.Execute<List<GithubPullRequestModel>>(request);
+
+                context.WriteLine($"Adding comment to issue {recentIssue.Number}");
```
abjerner: As we're listening for more than one repository, we should probably log some information about the repository as well. Eg. start the line with something like `[Umbraco-CMS/1234]`.

OurUmbraco/NotificationsCore/Notifications/ScheduleHangfireJobs.cs
```diff
@@ -52,6 +58,16 @@ public void UpdateMeetupStats()
             RecurringJob.AddOrUpdate(() => UpdateMeetupStatsJsonFile(), Cron.MinuteInterval(15));
         }
 
+        public void UpdateUpForGrabsIssues()
+        {
+            RecurringJob.AddOrUpdate(() => AddCommentToUpForGrabsIssues(), Cron.HourInterval(100));
```
abjerner: This will run each 100th hour. `Cron.MinuteInterval(100)` should make it every 100th minute instead.

PR #438, opened 2019-05-21
Added support for logging in and signing up with GitHub

abjerner:
With this PR, I've added logic so that visitors have a secondary login option, meaning that they can now log in using their GitHub account.

Pressing the **Login with GitHub** button works like this:

1. If a person already has an account on Our Umbraco, and have linked that account with a GitHub account, login should work right away

1. If a person already has an account on Our Umbraco, but hasn't yet linked with a GitHub account, the person will see an error about the email is already used by another user (assuming they're using the same email in both places). Solution is to log in with email and password, and then link with GitHub. Then **1)** should work from there on.

1. If a person doesn't already have an account on Our Umbraco, a new account will be created for them based on the name and the email address from their GitHub account. The new account will then be linked with their GitHub account, but don't have a password until the person manually adds this (not part of the signup process).

There are a few scenarios that will cause an error (which are handled in `LoginController.cs`). I'd like to test this part a bit more, so I've marked this PR as a draft for now. 

image


abjerner:
I couldn't help but play a bit around with the error messages. For the current implementation of this PR, an error message would look like this:

image

It's just a plain white page with the error message written in sans serif. If we create a "minimal" master page, we can use that to make some more friendly error message (inspired by the signup and forgotten password emails from Umbraco):

image

Like on the "full" version of Our, the logo is a link to the front page. I haven't committed this yet, but will probably to so tomorrow after some post-Codegarden sleep.

warrenbuckley:
Hi @abjerner 
What is the status of this PR - want do you need to do finish this off and anything I could to help assist at all?

abjerner:
@abjerner good question. I'll have a look 😉

abjerner:
@warrenbuckley I made some changes back in June that I hadn't committed to this PR, but to another branch. I've now merged that back to the branch used for this PR.

I also found an issue with the login controller preventing users from signup up with their GitHub account. I've fixed that as well.

This PR should now be ready to be merged, so I've removed the draft flag 🎉

If something fails during login or signup, an error message will be shown to the user. In some cases the error message shown will simply be:

> Oh noes! This really shouldn't happen. This is us, not you.

But then more information is appended to the Umbraco log. Mostly this covers errors that shouldn't happen, so if they actually turn out to happen more than once in a while, we can perhaps look into making the error messages more detailed for the user.

warrenbuckley:
@nul800sebastiaan @hartviglarsen do we have plans to roll this out at all?

hartviglarsen:
> @nul800sebastiaan @hartviglarsen do we have plans to roll this out at all?

It would make a lot of sense to roll this out. :) 

I'll find some time for review either this week or the beginning of the next.

Issue #439, opened 2019-05-25
Mobile navigation toggle is broken on mobile


bjarnef:
It seems the mobile navigation toggle is broken on mobile, but only in frontpage.

It is most likely a javascript error, note the hash in the url.

![Screenshot_20190525-123957_Chrome](https://user-images.githubusercontent.com/2919859/58368243-9fcd7500-7eea-11e9-8a04-70cc2e7949ab.jpg)

abjerner:
This is a weird one. I don't get any errors (when emulating mobile on desktop), but still experience the same as you describe.

I think #440 would fix it though 😉

bjarnef:
@abjerner it has happened after some of the recent changes.

bjarnef:
There is a similar issue with the quick menu doesn't open on desktop, when clicking the user/avatar on frontpage. It does work on subpages though.

![image](https://user-images.githubusercontent.com/2919859/58544594-ddd7db00-8201-11e9-8f80-56de3a765feb.png)

bjarnef:
and it seems the error in the Twitter partial are breaking part of the layout 🙈 

![image](https://user-images.githubusercontent.com/2919859/58544696-15468780-8202-11e9-8316-985902c59882.png)

nul800sebastiaan:
The Twitter error has been fixed now, I'll close this one :)

PR #440, opened 2019-05-25
Fixed mobile navigation issues on the front page (maybe)

abjerner:
As reported by in #439, the mobile navigation is broken on the live site. I can't however reproduce this locally with the newest code, but inspecting the code on the live site, it appears to be the same.

I played a bit around in the console of the browser on the live site, and placing the click event on the `a` instead of the `div` seems to fix this. But I can tell for sure as I can't reproduce it locally.

### DIFF
--- a/OurUmbraco.Client/src/js/app.js
+++ b/OurUmbraco.Client/src/js/app.js
@@ -22,7 +22,7 @@ jQuery(document).ready(function () {
     });
 
     // Mobile navigation
-    $("#toggle").click(function (e) {
+    $("#toggle a").click(function (e) {
         e.preventDefault();
         $(".cross").toggleClass("open");
         $("nav").toggleClass("open");


bjarnef:
@abjerner maybe there is something in Google Tag Manager or other external scripts since it works on all other pages except frontpage and you can't reproduce it locally.

abjerner:
The Twitter integration on the frontpage currently fails as well, injecting the full HTML of the error page. I bet this could cause the menu to fail as well.

https://our.umbraco.com/error/?aspxerrorpath=/umbraco/Surface/TwitterSearch/TwitterSearchResult

So maybe my fix isn't necessary if the Twitter error is fixed.

bjarnef:
Yes, og looks kinda weird in the footer at the moment 🙈

nul800sebastiaan:
I've fixed the Twitter feed (kinda, not sure why it's failing but it's now failing gracefully), but this won't hurt either, in case something else weird happens in the future. Thanks @abjerner !

PR #442, opened 2019-06-03
Twitter: Removed call to get authenticated user

abjerner:
Twitter has made some breaking changes, meaning that some JSON properties in the user object that used to be returned by their API are no longer there. This breaks Tweetinvi, and apparently the developer wont be able to release an update to address this for some months.

Our's current implementation of the Twitter feed makes a request to get information about the authenticated Twitter user. As Tweetinvi are no longer able to parse the response, the feed fails.

The solution is therefore just to remove the line, as apparently isn't used for anything. I tried to remove the line in my local setup, and the feed now works again.

### DIFF
--- a/OurUmbraco/Community/Twitter/TwitterService.cs
+++ b/OurUmbraco/Community/Twitter/TwitterService.cs
@@ -36,8 +36,6 @@ namespace OurUmbraco.Community.Twitter
                                     ConfigurationManager.AppSettings["twitterUserAccessToken"],
                                     ConfigurationManager.AppSettings["twitterUserAccessSecret"]);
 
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Ah this is great, thanks so much for getting to the bottom of this one! I think at some point we were under the impression that getting the authenticated user was the only way to authenticate, but this seems to work just fine.

Thanks again! 👍

bjarnef:
@nul800sebastiaan it seems the issue is back. Has this been deployed to our.umbraco.com or is it a different issue? The main issue is that in break the mobile navigation and the quick menu via avatar on desktop.

image

PR #445, opened 2019-06-11
ContributorBadgeService: improved the way we find members by their GitHub user IDs

abjerner:
If at first you don't succeed, try again 😄 

With PR, the service will look up a member a GitHub user ID with a very specific raw SQL query instead of using the member service, which has quite a bit of overhead.

The raw query will find the ID of the member, and the member service is then still used to get an instance of `IMember` representing that member.


nul800sebastiaan:
Looks good @abjerner - thanks! 👍

Issue #449, opened 2019-06-14
Browse most popular packages title is wrong


prjseal:
When you go to the packages page without any query string you see a title like this:

## Browse most popular packages

This is incorrect, because you are looking at the latest packages not the most popular, which can be misleading for people.

So it should look like this by default instead:

## Browse latest packages

I have created a PR which I will link to this issue.

nul800sebastiaan:
Thanks @prjseal ! Will have a look at it this week! 👍

prjseal:
Any joy with this one @nul800sebastiaan ?

nul800sebastiaan:
Fixed in https://github.com/umbraco/OurUmbraco/pull/450

PR #450, opened 2019-06-14
- fixed the issue with the headline showing as most popular packages

prjseal:
- fixed the issue with the headline showing as most popular packages by default when we are actually seeing latest packages.

So now it depending on the order mode it will show as 'latest', 'recently updated' or 'most popular'.

This fixes issue #449


nul800sebastiaan:
Thank you kindly @prjseal - sorry for the delay on this one, lovely work!

PR #460, opened 2019-08-14
Fix Twitter template

hartviglarsen:


nul800sebastiaan:
Nice one, thanks for the assist!

Issue #461, opened 2019-08-26
Analytics querystring messes up browser title


jannikanker:
I was reading the latest and greatest product update blog post and followed a link to the docs.

The link (https://our.umbraco.com/documentation/Reference/Mapping/?_ga=2.49479141.217517450.1566817370-1331147576.1557146764/#full-example) has some querystring params in it, which seem to mess with the browser title:
![image](https://user-images.githubusercontent.com/1769849/63687708-97e0e200-c805-11e9-939c-1c52cc08c1e9.png)

Checking the HTML source, the title is set to:

"?_Ga=2.49479141.217517450.1566817370 1331147576.1557146764, Reference - Our Umbraco"

Which I'm guessing isn't really what we want? :-)

nul800sebastiaan:
Yeah that looks a bit weird. 

Happy to look at a PR if you would like to  help out! 👍

umbrabot:
Hi @jannikanker,

We're writing to let you know that we've added the `Up For Grabs` label to your issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly PR team bot :-)

jannikanker:
I'll take a look :-)

PR #462, opened 2019-08-27
Potential fix for issue #461: Update DocumentationContentFinder.cs

jannikanker:
The code that sets "topicTitle" uses Request.RawUrl as a base for splitting the current page "path", reversing it and displaying it in <title />. Not sure if that's the best approach all round, but RawUrl includes the querystring which I'm sure sucks.

I've given up trying to run the website locally (everything and HTTPS seems to screw me over for more than an hour now) so this is **_untested_** (!!!), but I'm hoping someone (hi @nul800sebastiaan ;-) ) with a working local setup might be able to do a quick test to see if this oneliner fixes https://github.com/umbraco/OurUmbraco/issues/461 ? Pretty please?

### DIFF
--- a/OurUmbraco/Documentation/DocumentationContentFinder.cs
+++ b/OurUmbraco/Documentation/DocumentationContentFinder.cs
@@ -62,7 +62,7 @@ namespace OurUmbraco.Documentation
             // set the context vars
             var httpContext = contentRequest.RoutingContext.UmbracoContext.HttpContext;
             httpContext.Items[MarkdownLogic.MarkdownPathKey] = mdFilepath;
-            httpContext.Items["topicTitle"] = string.Join(" - ", httpContext.Request.RawUrl
+            httpContext.Items["topicTitle"] = string.Join(" - ", httpContext.Request.Url.AbsolutePath
                 .Split(new[] { '/' }, StringSplitOptions.RemoveEmptyEntries)
                 .Skip(1)
                 .Reverse());


abjerner:
I agree that `AbsolutePath` is what we want, so this looks correct (I haven't tested it either though).

What are you issues with getting the site up and running? The forms authentication will require you to run HTTPS, so this might be what you're experiencing. You can disable it by setting `requireSSL` to `false`:

https://github.com/umbraco/OurUmbraco/blob/master/OurUmbraco.Site/web.vsts.config#L330

jannikanker:
@abjerner Excellent :-)

I'm not even sure what the issue is. Now that I think of it, it could be as simple as Chrome caching some http<->https redirect stuff as I first ran the site with requireSSL="true" and failed. I'll check tomorrow.

abjerner:
@jannikanker Sounds plausible. I can recommend this Chrome extension for clearing cache (and 301 redirects) 😉

jannikanker:
> @jannikanker Sounds plausible. I can recommend this Chrome extension for clearing cache (and 301 redirects) 😉

@abjerner Which extension would that be? ;-)

jannikanker:
@abjerner Just spent a good hour again today, running into ALL THE ERRORS. Perhaps we could update the readme to include how to pull down the Markdown files required to have a working Documentation section? 😃 Would have loved to have known that beforehand.

Aaaaanyway, I've tested my little one liner now, and I can confirm that it keeps the browser title sane and sound, even if the URL has querystring parameters.

jannikanker:
Although... The readme says to rebuild documentationIndexer. But there's also a tab called "GitHub Documentation" with a "Get Docs!" button, which seems to do ... something more? Which one is correct? The latter worked perfectly for me.

nul800sebastiaan:
Sorry for the delay here @jannikanker ! This seems to work just fine.

The correct button.. well.. we don't really know either.. whichever one works first! :-o

Issue #469, opened 2019-09-10
Quick Menu View My Public Profile Link GitHub Error


prjseal:
If you have your GitHub account linked to your Our Umbraco account, the 'View my public profile' link in the quick menu goes to a 404 page.

![Quick menu view my public profile link](https://i.imgur.com/7LEdwXc.png)

![Page not found](https://imgur.com/T5zo8BH.png)

It would be better to use the same link that is used when you click on someone's name in a comment on the forum.

![Forum comment name url](https://imgur.com/lCWDgHd.png)

It then goes to the correct page for the member

![Correct public profile url](https://imgur.com/FG6SRYa.png)

I have already created a fix for this and will link it to this issue.

abjerner:
Hi @prjseal,

I'm a bit unsure why the `@` is included in your username. It doesn't do that when I link with my own GitHub account. I think it would be better to fix this part so we can still use vanity URLs based on the GitHub username.

What does it say if you go to the [**Edit profile** page](https://our.umbraco.com/member/profile/)? Mine looks like this:

![image](https://user-images.githubusercontent.com/3634580/64603201-4bc69d80-d3c0-11e9-9785-e6ccce3ad9a1.png)

prjseal:
Hi @abjerner 

This is how it looks on mine:

![paul seal github linked account](https://imgur.com/8cxRjQM.png)

abjerner:
@prjseal did you possibly write your GitHub username manually before linking? That could explain why it has a `@` (I think).

abjerner:
And what happens if you unlink and link again?

prjseal:
That has worked for me. I linking and then linking again. I must have added my GitHub account name before the linking with GitHub was added to the site.

abjerner:
@nul800sebastiaan would it make sense to create a job that fixes the usernames if there are anyone left with `@` in the username? I don't really now how many users we're talking about.

@prjseal Depending on what Seb answers, it might make sense to update your PR so `@Url.GetProfileUrl(profile)` strips any leading `@`.

nul800sebastiaan:
I don't really want to run this as a job, it would be better to just strip any leading `@` signs from the profile link while building it.

abjerner:
@nul800sebastiaan I think you may have to, but I'll dig more into the issue.

If the property does contain a `@`, we can strip it when rendering the profile link. But the username is also used for vanity URLs, which may cause problems (I'll have to test that) when searching for the username.

[...]

PR #470, opened 2019-09-10
Fix 'view my public profile' link in the quick menu

prjseal:
This PR fixes the issue #469 'View my public profile' link which occurs when the member has linked their GitHub account to their our account.

Quick menu view my public profile link

Page not found

This fix uses the same link that is used when you click on someone's name in a comment on the forum.

Forum comment name url

It then goes to the correct page for the member

Correct public profile url

### DIFF
--- a/OurUmbraco.Site/Views/Partials/Members/QuickMenu.cshtml
+++ b/OurUmbraco.Site/Views/Partials/Members/QuickMenu.cshtml
@@ -64,7 +64,7 @@
                             </span>
                         }
                         <div class="public-profile">
-                            <a href="@Url.GetProfileUrl(profile)">View my public profile</a>
+                            <a href="/member/@profile.Id">View my public profile</a>
                         </div>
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Thanks @prjseal - and sorry for the delay on this one. There must be a different problem, since your github username makes a "vanity URL" for you. So in this case, it should have been: https://our.umbraco.com/members/prjseal

So we'd need to figure out where the additional @ prefix is coming from. Any ideas?

abjerner:
@nul800sebastiaan Paul and I had a chat about that in https://github.com/umbraco/OurUmbraco/issues/469 - it was most likely because Poul had entered the GitHub username manually before auto-linking was supported.

It's also still possible to manually specify your GitHub username when signing up. I think it would be best if this part was removed.

nul800sebastiaan:
I'll close this one for now then!

PR #482, opened 2019-10-01
Adding other repos

crgrieve:
Adding RFC and organisers repo to github service.

Is there any others that should be included in contributions?

PR #489, opened 2019-10-14
updating repos list on page

crgrieve:
Hello,

I realised on PR #482 I had updated the logic but not the list on homepage of which repos are included for contributions, so I have added them now :)

Should look like this:
image

Thanks,
Carole


crgrieve:
Thanks @abjerner.

oops, I copied them from the first one in existing file, it had the missing closing quote on the blank attribute and it copied over to those! Fixed now.

Thanks,
Carole

abjerner:
@crgrieve no worries. But I'm not seeing the fix here on GitHub. Have you pushed it yet?

crgrieve:
That should be it now 😇

hartviglarsen:
Thanks @crgrieve for the PR and @abjerner for the comments. :)

Yeah.. ideally we should use the `GithubService`. I'll have that updated at some point.

PR #490, opened 2019-10-15
Cleanup release URLs

callumbwhyte:
I went down a bit of a rabbit hole, sorry...

It was really bugging me that the URL for each version on the ["all releases" page](https://our.umbraco.com/download/releases/) contained a double slash...

And then I noticed that some URLs elsewhere on Our point to the old `/contribute/releases/{version}` URL instead, so I decided a centeralised helper would be a nice way to fix this


nul800sebastiaan:
Nice one @callumbwhyte! Really happy with a bit of cleanup too, thanks for that! ⭐️

PR #497, opened 2019-10-25
Added favicon.ico because YES PLEASE

abjerner:
Our doesn't have a `favicon.ico` file, so I've added it. The main Our Umbraco site specifies that the following PNG file should be used as the favicon:

https://github.com/umbraco/OurUmbraco/blob/master/OurUmbraco.Site/Assets/images/app-icons/favicon.png

This works like it should on the main site, but other parts of the domain - eg. the API documentation - then lacks a favicon. By adding a `favicon.ico` file to the root of the site, the documentation and other parts of Our will automatically pickup this favicon 😍 

The added file is just a copy of the `favicon.png` mentioned above. The browsers doesn't care that it's actually a PNG file 😄 

image


nul800sebastiaan:
YES PLEASE! 😁

PR #508, opened 2019-11-07
Nuget package downloads

dawoe:
One of things we like to achieve with the new package team is to show nuget downloads on our. (https://github.com/umbraco/Umbraco.Packages/issues/9)

This PR is the first step in making this possible. It's the scheduled job to get the nuget downloads for packages that have a nuget url defined.

It would be nice if we could merge this. So we can build on this in future PR.

Dave


abjerner:
...but it's spelled NuGet 🙃 

Anyways, do we have an overview over how many packages we're talking about? The `ImportNugetPackageDownloads` method may hit the NuGet API pretty hard. And with your proposed code, it runs every hour. Perhaps this should run this in bathes or a sleep should be added somewhere?

But then on the other hand, one could argue that NuGet/Microsoft would have the infrastructure to handle this.

Btw - NuGet themselves (at least on the website, I'm not sure about the API) only update download stats once a day or a few times each day, so maybe the job should only run once a day as well or each 12th hour.

dawoe:
Hi @abjerner 

They have a rate limit of 3000 requests per minute. That is why I handle them in batches of 5 packages in one query. So in theory we can handle 15000 packages per minute.

For now I think that will be good enough. 

Dave

dawoe:
We will also need to investigate if the Nuget API supports getting data from a specified time period. Now you can filter on popular packages on our and that uses the number of downloads over a set period of time.

dawoe:
Added new commit which indexes nuget downloads in the project index. It will combine our and nuget downloads

dawoe:
So this PR is basically ready to show the nuget downloads on the project detail page and total amount of downloads in the project listing. 

What doesn't work is taking new nuget pacakges into account when searching for most popular packages. But I wonder if that is still working correctly at the moment. If I go to our and look at the most popular packages I see some in there that makes it hard to believe they are among the most popular these days :
https://our.umbraco.com/packages?orderBy=popularity

For example I see Courier, uComponents and uWebshop among the most popular packages on the first page.

jmayntzhusen:
You can find the popuularity score calculations here: https://github.com/umbraco/OurUmbraco/blob/master/OurUmbraco/Our/Examine/ProjectPopularityPoints.cs#L112 

This is another thing that could probably do with an update 🙂

jmayntzhusen:
To provide a bit of context, currently Our on live has 1345 packages in the `projectIndexer`, so I'm sure the rate limit won't be a big issue for now 🙂

dawoe:
I have done some research in getting download stats from the nuget API is not possible.
So the only options is to keep track of the history our selves (in a database table).

If we have that in place we could than update the search on popular packages to take this into account.

How is creating of custom tables handled in our ? I would expect by using migrations, but I can't find them in the source anywhere.

Dave

warrenbuckley:
@dawoe not sure if you seen this API endpoint from Nuget Docs

### Examples
https://azuresearch-usnc.nuget.org/query?q=PackageId:UmbracoCms
https://azuresearch-usnc.nuget.org/query?q=PackageId:Our.Umbraco.Nexu

### Further Docs
https://docs.microsoft.com/en-gb/nuget/api/search-query-service-resource#request-parameters
https://docs.microsoft.com/en-gb/nuget/consume-packages/finding-and-choosing-packages#search-syntax

hartviglarsen:
@dawoe Database table creation is to my knowledge handled in this file: https://github.com/umbraco/OurUmbraco/blob/master/OurUmbraco/Our/MigrationsHandler.cs

See CreateGitHubAutoRepliesTable method for an example.

dawoe:
@warrenbuckley yes I have seen this. But I meant getting downloads over a period of time..aka last 6 months.

@hartviglarsen thanks for the pointer

Dave

warrenbuckley:
Why do you want it over a period of time?! I am probably missing something super obvious?!

jmayntzhusen:
@dawoe 

Maybe it could just be # of downloads on the current version instead? The nuget api has that available 🙂 Guess it may punish "rapid development" a bit though 🤔

dawoe:
@warrenbuckley we need to get the downloads over time for the search on popular packages to take nuget downloads into account. Which is broken now anyways. Can't imagine that uComponents and uWebshop are still popular

But maybe this PR can get merged...and do these improvements later on

dawoe:
This PR is now ready to be merged.

When retreiving downloads stats from nuget I get the total downloads. I also get the first published date on nuget to see how long the package exists there and calculate the average number of downloads per day since the package has been released.

This number is later used to calculate the package popularity.

jmayntzhusen:
Hey @dawoe 

This is great, and works like a charm!

There is one little thing I'd like changed though. In the popularity rankings we don't check for the `isRetired` field on package nodes.

image

I know that wasn't done before either, but with the new check on daily NuGet downloads it will impact a few packages that are listed as retired on Our. 

Could you check on that field and then not include them in the popularity rankings if it's true? 🙂

dawoe:
Hi @jmayntzhusen 

The search on popular package now excludes the retired packages. You can test it by rebuilding the project index. And then doing the popular packages search. You will not see nested content anymore.

Dave

hartviglarsen:
You all freaking rock!

Thanks to @dawoe for putting in the hours to make this and thank the rest of you for your constant feedback/comments.

We're so blessed to have a community like this. #H5YR

PR #510, opened 2019-11-07
Remove download requirement from voting on compatibility

jmayntzhusen:
We've had a talk at the London hackathon - it doesn't make sense to require downloading the package before you can report compatibility as it means noone really does it.

Seems the method was only ever used here, so cleaned up the model as well.

https://github.com/umbraco/Umbraco.Packages/issues/20


hartviglarsen:
Nice, Jesper (and Lotte!). This makes total sense. :)

emmaburstow:
👍

PR #511, opened 2019-11-07
Publish Packages with APIKey (JWT Token)

warrenbuckley:


OurUmbraco/Auth/UmbracoAuthTokenDbHelper.cs
```diff
@@ -0,0 +1,94 @@
+﻿using Umbraco.Core;
+using Umbraco.Core.Persistence;
+
+namespace OurUmbraco.Auth
+{
+    public static class UmbracoAuthTokenDbHelper
```
Shazwazza: Why are we using statics? Even though the codebase for our might not be that wonderful, we should refrain from making it any worse. All new code should be using DI.

warrenbuckley: It was quick & dirty to get a POC working & was copied/pasted from super old code I had for a JWT implementation.

Will see if I can get time to update this today/tomorrow

However I upgraded the JWT nuget dependency package to latest which requires a newer version of .NET framework. I think @nul800sebastiaan said the VM running Our may not have latest frameworks available.

nul800sebastiaan: It doesn't and I won't be able to install framework 4.8 either.

nul800sebastiaan: For now, we're stuck on `.NET Framework 4.5.2` as confirmed through https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/how-to-determine-which-versions-are-installed

warrenbuckley: OK thanks will see if I can fix this PR to get the JWT library to work with 4.5.2

warrenbuckley:
@Shazwazza I have removed the statics in this PR now. The problem I have is the JWT library will need to be like V3 or something for it to work .NET Framework 4.7.2 that is on the VM/Server for Our.Umb

jmayntzhusen:
I reduced the JWT version to 3.0.1 which is the lowest compatible version with .NET 4.5.2, and then had to redo a few of the JWT token building parts to work in that version.

Have tested it locally with the package tool and it can not send again.

What we need for now:

- We still need @Shazwazza to do a security review of the whole auth setup here
- We will probably want to add some sort of functionality to set the package you push as current, set the compatible versions & .NET versions, and also maybe have a flag to archive all other versions?

jmayntzhusen:
I rebased this commit on Master and Github has a hard time handling it for some reason.. It includes a bunch of changes that are already in master here.. https://github.com/umbraco/OurUmbraco/compare/package-tokens shows the real picute..

PR #512, opened 2019-11-07
Update the Australian prevalue to actually be correct.

hartviglarsen:
Changes _Australia_ to _Oceania_ when creating new banners in the Backoffice.


nul800sebastiaan:
I've just updated this manually on the live server, seems a bit easier ;-)

PR #513, opened 2019-11-11
Upgrade to .NET FX 4.8.0

warrenbuckley:
Sets both CSProjs to NET FX 4.8

### Note
Please test this out locally or on a UAT environment to ensure nothing gets messed up. For my local instance I had to do a fresh checkout

## Why
Because the JWT library requires at least 4.7.x and shipping an older library for something that is security related feels a bit wrong

This current WIP/draft PR for adding JWT API tokens for reference https://github.com/umbraco/OurUmbraco/pull/511


nul800sebastiaan:
Unfortunately we can't install .NET Framework 4.8 on the Our at this point, so we'll need to look for alternatives 🙈

PR #519, opened 2019-11-19
Remove Github contribute redirect

hartviglarsen:
**What**
This PR removes the redirect to Github on /contribute.

**Why?**
We will be creating a new one (contribute page) and therefore need the ability to use that path. :)


nul800sebastiaan:
This has been changed a bit and that page is now called "get involved" so we won't need to remove this any more (hopefully also preserves existing links to `/contribute`) to still point to github.

PR #524, opened 2019-12-10
updates in line with Umb 10th Dec security advisory workaround

liamlaverty:


nul800sebastiaan:
Ah I just saw this, thanks very much for the help @liamlaverty! I was already in the process of updating to the latest version so we don't need to use a workaround. The upgrade was rolled out have an hour ago. Great minds think alike! ;)

Thanks again!

PR #530, opened 2020-01-09
Package page banner

jmayntzhusen:
Had some requests to not automatically redirect to the about packages page from the top menu.

Now it will redirect to the package overview as it used to, but have a banner at the top that links to the about page:
image

The style of the banner is the same as on https://our.umbraco.com/community/ and I have added a bit to the styling for it, but nothing that will affect the current banner as it doesn't contain <div> or <a> elements 🙂


abjerner:
@jmayntzhusen great change 👍

hartviglarsen:
This is awesome!

As I mentioned earlier I'll do my very best to take a look at it later today if I can. Otherwise I'll do it tomorrow morning :)

#H5YR

PR #535, opened 2020-01-31
Package stats

jmayntzhusen:
@nul800sebastiaan 

This implementation is not caching the db calls - it does 3 calls per page hit and the page is only for hq. Should it be cached?

I also left in the old GetPackageStatisticsData() method, but am not using it - should it just be removed? 🙂


hartviglarsen:
Is this for public or "internal" use? In case of the latter I don't see any harm in it not being cached. We're not going to be looking at it 25 times per day. 🙂

jmayntzhusen:
This is supposed to be an internal page for now, where we can expand upon what we track.

If stats are deemed relevant for other things (fx a top x most downloaded packages past month could make sense in the package overview) we will consider doing it differently for that later.

nul800sebastiaan:
> I also left in the old GetPackageStatisticsData() method, but am not using it - should it just be removed? 🙂

Yep just remove it.

Depending on how fast the queries return now you might want to consider caching. I suspect you won't need it for now though.

jmayntzhusen:
Alright I cleaned it up for now 🙂

PR #539, opened 2020-02-07
Get number of downloads from examine index instead of db, so nuget do…

### LINKED ISSUE #537: Update member project download counts to include NuGet dls.
On the project overview page (https://github.com/umbraco/OurUmbraco/blob/master/OurUmbraco.Site/Views/Partials/Projects/ListProjects.cshtml) we show the download count like this (simplified):

```csharp
var result = search.SearchResults.Skip((page - 1) * pageSize).Take(pageSize);

@foreach (var childPage in result)
{
    @GetField(childPage.Fields, "downloads", "0")<span><i class="icon-Download-alt"></i></span>
}
```

So it is the `downloads` field from the projectIndex. 

On the member view page (https://github.com/umbraco/OurUmbraco/blob/master/OurUmbraco.Site/Views/Partials/Projects/MyProjects.cshtml) we get it like this:

```csharp
@helper RenderProject(IListingItem project)
{
    @project.Downloads<span><i class="icon-Download-alt"></i></span>
}
```

Where project gets its downloads from here: https://github.com/umbraco/OurUmbraco/blob/master/OurUmbraco/MarketPlace/NodeListing/NodeListingProvider.cs#L88-L106 where it ends up with a db call.

To allow the full download count on member pages we will need to get them from the index instead.

dawoe:
Get number of downloads from examine index instead of db, so nuget do…wnloads are included

Fixes #537

### DIFF
--- a/OurUmbraco/MarketPlace/NodeListing/NodeListingProvider.cs
+++ b/OurUmbraco/MarketPlace/NodeListing/NodeListingProvider.cs
@@ -3,10 +3,12 @@ using System.Collections.Generic;
 using System.Configuration;
 using System.Linq;
 using Examine;
+using Examine.SearchCriteria;
 using OurUmbraco.MarketPlace.Extensions;
 using OurUmbraco.MarketPlace.Interfaces;
 using OurUmbraco.MarketPlace.Providers;
 using OurUmbraco.Our;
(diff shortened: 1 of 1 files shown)


jmayntzhusen:
Works like a charm 🎉👍

PR #541, opened 2020-02-07
Update Documentation.cshtml

erikjanwestendorp:
Deleted v7 Getting started section on home


jmayntzhusen:
Good catch @erikjanwestendorp 

For anyone wondering, the spot he is removing `v7` from is on the homepage of Our:
image

Thanks for taking the time to contribute 👍🎉

PR #542, opened 2020-02-08
Allow order by downloads

jmayntzhusen:
Allows users to order by downloads as well as popularity and last update:
image

Additionally added a little indicator of how many packages are in the search results as that was never apparent before.

### DIFF
--- a/OurUmbraco.Site/Views/Partials/Projects/ListProjects.cshtml
+++ b/OurUmbraco.Site/Views/Partials/Projects/ListProjects.cshtml
@@ -165,7 +165,7 @@
     }
 }
 <h1>Browse @headline packages</h1>
-<p><em>A total of @total packages match your current criteria</em></p>
+<p><em>@( total == 1 ? $"{total} package found" : $"{total} packages found")</em></p>
 <script type="text/template" class="search-item-project">
     <div class="box">
         <div class="row">
--- a/OurUmbraco.Site/Views/Partials/Projects/ListProjects.cshtml
+++ b/OurUmbraco.Site/Views/Partials/Projects/ListProjects.cshtml
@@ -22,6 +22,10 @@
     {
         headline = "recently updated";
     }
+    else if (orderMode == "downloads")
+    {
+        headline = "most downloaded";
+    }
 
     if (version != null)
     {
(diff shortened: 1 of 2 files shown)


OurUmbraco.Site/Views/Partials/Projects/ListProjects.cshtml
```diff
@@ -158,6 +165,7 @@
     }
 }
 <h1>Browse @headline packages</h1>
+<p><em>A total of @total packages match your current criteria</em></p>
```
LottePitcher: If there was only 1 package that matched the search criteria, this would display as "1 packages".  Would be nice if it displayed as "1 package" in that scenario.

LottePitcher:
Great idea @jmayntzhusen to add this!  

If the user hasn't entered any search criteria, as per your screenshot, could the 'total' message be tailored to remove the reference to 'search criteria' (as the user hasn't entered any)?  As it's actually the total number of public packages on Our isn't it?

LottePitcher:
Ok I had a bit more of a play with the current search page.  The list of packages is shown initially if the user has clicked the 'Packages' top navigation link. In that scenario they have not entered any criteria. The side navigation links allows them to change the sort order, or filter by category/version.  They could also filter the search results by specifying criteria in the search box.  As it's a bit of a mix of criteria/ordering maybe it's easier to have a simple 'X listed' message somewhere. Perhaps it could positioned to the right of the H1 above the versioning? Then wouldn't have to work out how to make the message relevant to what the user is actually seeing.

I have not really noticed before that the search box on the packages page does client-side updates of the search results once you've typed more than 3 characters in the box. If you press Return whilst in the search box it redirects you to the full Our search results page.  Seems a bit of a shame to 'lose' them from the packages part of Our. Appreciate this is slightly off topic for this PR but do you think this is something that could/should be changed too?

jmayntzhusen:
Good points @LottePitcher ...

Hmm what if the message was simply 
```
@total package(s) found
```

No need to mention criteria, etc. Then it would work for 0, 1, .. n packages and would then make sense if you specify it by version, etc then it just finds less 🤔

As for the search stuff, feel free to submit a PR but think it's a bit out of scope of this one 😁

LottePitcher:
@jmayntzhusen yes agreed, search box 'enter' key press behaviour definitely out of scope here!

Good idea to keep message simple as you suggest. Still think it would be neater if it showed

```1 package found```
or
```X packages found```

It only needs to differ ```if (total == 1)``` ...

jmayntzhusen:
Made a change so it will now look like this:
image

but will say `package` if there is exactly 1

jmayntzhusen:
Closing this for now, a few other larger changes has happened, and will probably have a separate task in the future to determine what sort of filtering and sorting makes sense to add 🙂

PR #551, opened 2020-02-21
Grid based packages homepage.

KevinJump:
A Draft layout of the packages homepage that puts the package listings into a grid. 

I think i need some feedback as to if this is worth tidying it up, etc before finishing the PR.

This PR doesn't change any of the functionality for the packages section, it contains layout and styling changes only.

image

- Package Homepage, has two listings one for popular and one for latest, (both show first 10 results)
- any subsequent page shows 30 results and pagination,
- the drop downs on the top right allow you to filter by category / umbraco version

### WIP: 
- Not mobile ready,
- Not sure about spacing units across the page.
- Not sure about headings for each variation.


KevinJump:
Added sort order to the filter .

image

not convinced it isn't starting to look a little cluttered.

LottePitcher:
@KevinJump is the title ('Backoffice Extensions For Umbraco 8' in your last screenshot) completely derivable from the selections made in the dropdowns and the search box above?  As if so if there an argument for removing that title entirely and have the filter dropdowns left aligned?

jmayntzhusen:
@KevinJump 

I know you are only looking into UI atm, and that is fine, but it would be awesome if we could also get rid of all the Code duplication that we have on this page vs the API that sends the feed into the backoffice. 
We could then ensure the same - or atleast a very similar UI was available for both the backoffice and Our.

If it is something you feel like taking a look at as part of this PR then the API for the backoffice is found through https://github.com/umbraco/OurUmbraco/blob/master/OurUmbraco/Repository/Controllers/PackageRepositoryController.cs#L79 

Otherwise it is something we should create an issue on. This is going to give us a lot of issues down the road I think. It is also the reason I haven't moved forwards with https://github.com/umbraco/OurUmbraco/pull/542 yet and why the new NugetDownloads update doesn't apply in the backoffice.

jmayntzhusen:
Hmm I can't seem to make pr's to your branch @KevinJump 

Fixed a bit of CSS for mobile though.

I think that we should get rid of the categories at the bottom of the page - they don't serve a lot of purpose there, especially with them in the top filtering as well 🙂

KevinJump:
don't think i've blocked anything on the repo :( - feel free to DM me the update and i will put it in now.

jmayntzhusen:
I could update on Github, just couldn't push from local 🤷‍♂️

With the small changes in https://github.com/umbraco/OurUmbraco/pull/551/commits/c426118b4d8b4ab4d7e4663076dc5d422ed3d911 it should look like this:

image

KevinJump:
Some more - to merge the mobile stuff with the slightly bigger filter boxes,

I will take a look at merging the search and backoffice search in another PR.

jmayntzhusen:
This is awesome @KevinJump !

PR #552, opened 2020-02-24
Publish Packages with APIKey

warrenbuckley:
This PR consists of several parts:

# 1: Functionality to make API Keys on Our

**Major parts**:
1. Added a migration to create a new db table for API Keys: https://github.com/umbraco/OurUmbraco/pull/552/files#diff-0e98eb2eb785d4572d0b2823827f93dfR3034
2. Added a new service to create, get and delete keys: https://github.com/umbraco/OurUmbraco/pull/552/files#diff-4e5283c8324db3e3bdef039bf04fc0dbR12

# 2: Public API + auth for updating packages from CLI tool

**Major parts**:
1. There is a new public api controller that has methods to get a list of package files for a project or to add a new zip file to a project: https://github.com/umbraco/OurUmbraco/pull/552/files#diff-43d4e00ead6d8da05d6b4d675acd9f38R14
2. The authentication requirement to access that controller is something Shannon has helped set up in https://github.com/umbraco/OurUmbraco/pull/560 - the main parts are in OurUmbraco/Auth/AppBuilderExtensions.cs & OurUmbraco/Auth/HMACAuthentication.cs

# 3: UI for managing API Keys

**Major parts**:
1. Migration to add a new node + macro to enable us to add a new page for members managing their packages: https://github.com/umbraco/OurUmbraco/pull/552/files#diff-0e98eb2eb785d4572d0b2823827f93dfR3064
2. New macropartial view that is used to manage API Keys, does so via a new API controller talking to the key service: https://github.com/umbraco/OurUmbraco/pull/552/files#diff-08387b760436250cecf94827709af656R12

# Testing

1. The migrations can be tested on a fresh pull of the solution, and check if the `projectAuthKeys` table exists, and that the Community -> Member -> Profile -> Packages -> API Keys node is there.
2. The key generation part can be tested as shown in the gifs on https://github.com/umbraco/OurUmbraco/pull/561
3. The auth parts can be tested by trying to push a package after generating the key - you will need to clone down https://github.com/umbraco/Umbraco-Packager-CLI and you can then run the tool locally with these commands:
```
umbracopackage push -k keyhere C:\Path\To\Your\Package
```

Note: The package has to contain a valid package.xml file to work 🙂


scottbrady91:
Hiya, I was pointed here from the Umbraco Collaborators Slack channel, and I had a few questions.

1. Could you explain the purpose of the JWT for me? What is it being used for, and by whom? 
2. Is there a reason you are storing JWTs in the database? It looks like you are treating it as an API key still. I assume that is why you are also decoding the token as opposed to validating it. But in that case, why sign the token, or why use JWTs at all?
3. Is there a reason symmetric signatures are used?
4. Is there a reason you are not using the Microsoft.IdentityModel libraries for JWT creation & validation?
5. There's a comment in UmbracoMemberAuthToken that refers to Basic auth. Is that being used somewhere?
6. Could you share an example JWT created by this code?

jmayntzhusen:
Hey @scottbrady91 

I will answer these as well as I can and hope @warrenbuckley can pitch in with the rest!

1. We will be launching a new CLI tool for packing and deploying Umbraco Zip packages. This token is intended to be a key you pass to the tool to upload new versions of your package directly to Our. As such the keys are specific to each package.

2., 3., 4., 5., I hope @warrenbuckley has answers for these!

6. The JWT token is partially based on an app setting, which locally will be `#{PackageAuthToken}#` but on the site the buildserver uses a very long auto generated key, the example here is from the string above, not sure if it does a difference.
```
eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJkYXRlX2NyZWF0ZWQiOiIyMDIwLTAyLTI0IDEzOjE4OjQzWiIsIm1lbWJlcl9pZCI6NDU3NiwicHJvamVjdF9pZCI6MjgwNjcxfQ.kCl8VRHLnkCZvjJ26nj59XgNzLaLKCaXjZMKCV11tdc
```

warrenbuckley:
Hi @scottbrady91 
I am glad we got someone like yourself who is more of an expert in this field than myself, so kudos to the power of the hive mind in open source 💪 

To summarise quickly before answering your questions in detail. I had learned about JWTs a while back and created a project to use JWTs with Umbraco, where most of this legacy code/work has been lifted & shifted from that was made into a proof of concept & thus has been picked up by Jesper to fine-tune this.

2: Yes as you say they are being stored in the DB as a way of an API key. The JWT payload contains the member ID and the project ID and was a way to know who this key belonged to.

As the original POC was made in a day due to me borrowing code from my old JWT project to help kickstart & move this project forward of having a .NET Core CLI tool to upload/publish packages with some kind of auth. It was quicker for me to re-use something I had than write something new to demo this POC. If you have suggestions on what we should be doing to help solve this problem then myself & Jesper will be happy to hear how it should be done 'correctly'


3: Shows how much I really know about the topic 🙈 as I have no idea what you mean by that


4: No real reason I did not use it, but I think at the time when doing my work back in 2015 the JWT.io site had listed this other C# NuGet package & was simple to use and understand


5: Can you point me to this, please. As I can't see any mention of **basic** in the code comments?!

OurUmbraco/Auth/UmbracoMemberAuthToken.cs
```diff
@@ -0,0 +1,146 @@
...
+            projectId = int.MinValue;
+
+            //If no Auth header sent or the scheme is not bearer aka TOKEN
+            if (ah == null || ah.Scheme.ToLower() != "bearer")
+            {
+                //Return null (by returning null, base method above will return it as HTTP 401)
+                return null;
+            }
+
+            //Get the JWT token from auth HTTP header param  param (Base64 encoded - username:password)
```
scottbrady91: Reference to Basic auth?

scottbrady91:
Is it only Our who is generating and receiving the JWTs?

jmayntzhusen:
> Is it only Our who is generating and validating the JWTs?

Yes we generate it on Our, output it on a member view then they can copy the token and use in a cli tool to authenticate their deployment of a file to Our.

scottbrady91:
Okay, thanks for all your answers!

Personally, I do not see any reason that this needs to be a JWT.  It's just an API key, which can happily be a random string.

Since you are storing the key in a database, you could just as easily store the member ID and project ID alongside it instead in the structured JWT (it looks like you already are storing these). The JWT isn't being validated, you're just checking it is in the database, so signing it isn't adding anything. JWTs are also typically short-lived, but these keys do not have an expiry.

Unless there is a requirement for multiple parties to start reading & validating the API keys, I don't see any benefit in the API key being a JWT.

Either way, the API key within the database should ideally be hashed. Display it to the user once, and then store it in a hashed format. A single round of SHA256 is fine, assuming the key is machine-generated and has decent entropy.

I would also recommend that you consider adding an expiry to these API keys.

jmayntzhusen:
Thanks @scottbrady91 !

In that case I suggest we just get rid of the external dependency on JWT and generate our own API keys. Will get to that as soon as I can!

Can you elaborate on why the API keys should expire? 🙂

scottbrady91:
API key expiry is a tricky one. A perpetual API key can pose a security risk but, on the other hand, an expired key can be a pain to update and negates some of the automation benefits. 

Since you are aiming for automation, I like the idea of expiring keys that have been unused for a few months (a year?) or allowing a user to set an expiration on the key.

In the same vein, a revocation mechanism is also important so that users can revoke old or compromised keys.

OurUmbraco/Auth/UmbracoMemberAuthToken.cs
```diff
@@ -0,0 +1,146 @@
...
+using System.Web.Http.Controllers;
+using System.Web.Http.Filters;
+using Umbraco.Core;
+using Umbraco.Core.Models;
+using Umbraco.Web;
+using JWT;
+
+namespace OurUmbraco.Auth
+{
+    public class UmbracoMemberAuthToken : ActionFilterAttribute
```
Shazwazza: Authorization and Authentication are 2 different things. This is combining both of them. Although that will work it's not the greatest way to do things. The authentication part should be in the very beginning of the request and add claims to the user/principal created in the request, the Authorization filter (this thing) should just be checking claims. This filter should also be an AuthorizationFilter not just an ActionFilter. 

I'm going to help review all of this and see if we can wrangle this in properly. Aspnet comes with all of the code and ability to Authenticate properly with jwt or bearer token authentication without having to re-invent the wheel since all of this code is manually re-doing what has already been done.

OurUmbraco/Auth/UmbracoMemberAuthApiController.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// When a user has been authorised with a JWT Auth token
+        /// This is the member represents the authorised member
+        /// that can then be used in the API controller for Umbraco API Service Calls 
+        /// such as creating content
+        /// </summary>
+        public IMember AuthorisedMember
+        {
+            get
+            {
+                return ControllerContext.Request.Properties["umbraco-member"] as IMember;
```
Shazwazza: See my other comment about authentication and authorization, this shouldn't be necessary since this is a hack instead of using Claims in the current user/principal in the request. This whole controller should be able to be removed if Claims are used. I'll see what i can do

OurUmbraco/Auth/UmbracoAuthTokenFactory.cs
```diff
@@ -0,0 +1,74 @@
...
+
+namespace OurUmbraco.Auth
+{
+    public class UmbracoAuthTokenFactory
+    {
+        private string _secretKey;
+
+        public UmbracoAuthTokenFactory()
+        {
+            _secretKey = new UmbracoAuthTokenSecret().GetSecret();
```
Shazwazza: This should all be using DI

scottbrady91:
Sure, you could use the OAuth client credentials flow for this but that wouldn't offer that many benefits for this automation approach and adds complexity. I'm not usually one to discourage the use of OAuth, but I'm not seeing the befit for your requirements.

For your concern around the same string being repeated over requests, the same would be the case if you used OAuth client credentials with a shared secret. If you have similar requirements to nuget, then the API key is not going to be in constant use and I would see an access token only being used once. Sure, you could use some of the certificate based client authentication mechanisms, but unfortunately, that's can be considered hard to use.

A nonce value wouldn't really apply here. From what I've been explained, this is machine to machine, not user interactive and does not include OpenID Connect.

For your requirements, I'm just struggling to see how the features of JWTs will benefit you. The same goes for the added complexity of OAuth.

If you are set on using JWTs in some form, then I highly recommend you stick to specifications such as OAuth and the JWT profile that IdentityServer implements.

OurUmbraco/Auth/UmbracoAuthTokenDbHelper.cs
```diff
@@ -0,0 +1,96 @@
+﻿using System.Linq;
+using Umbraco.Core;
+using Umbraco.Core.Persistence;
+
+namespace OurUmbraco.Auth
+{
+    public class UmbracoAuthTokenDbHelper
+    {
+        private readonly UmbracoDatabase Database = ApplicationContext.Current.DatabaseContext.Database;
```
Shazwazza: IMPORTANT! be sure to never do this since this means you can potentially pin a Database instance statically, for example if an instance of `UmbracoAuthTokenDbHelper` was made static. ALWAYS resolve a database instance from the DatabaseContext, never pin a database instance to a field unless you are absolutely sure that it's a transient instance.

Shazwazza:
@jmayntzhusen + @scottbrady91  I have pushed a new PR for this. There's lots of notes too https://github.com/umbraco/OurUmbraco/pull/560 let me know what you think and any questions, etc...

OurUmbraco.Site/Views/MacroPartials/Project/ApiKeys.cshtml
```diff
@@ -0,0 +1,77 @@
...
+    var projectId = Request["Id"];
+    int id = 0;
+}
+
+@if (!string.IsNullOrEmpty(projectId) && int.TryParse(projectId, out id))
+{
+    var project = Umbraco.TypedContent(id);
+    var member = Members.GetCurrentMember();
+
+    if (project.GetPropertyValue<int>("owner") != member.Id)
```
nul800sebastiaan, line 14: Can only owners create keys, collaborators are not allowed at this time?

OurUmbraco/Auth/AppBuilderExtensions.cs
```diff
@@ -0,0 +1,130 @@
...
+                    //to the request, this is the "User" that will get assigned to
+                    //the request with Claims.
+                    //If the token is invalid, then don't assign a ticket and OWIN
+                    //will take care of the rest (not authenticated)
+                    OnReceive = context =>
+                    {
+                        var requestPath = context.Request.Uri.CleanPathAndQuery();
+                        if (!TryGetHeaderValue(context, ProjectAuthConstants.ProjectIdHeader, out var projectId))
+                            throw new InvalidOperationException("No project Id found in request"); // this will never happen
+                        if (!TryGetHeaderValue(context, ProjectAuthConstants.MemberIdHeader, out var memberId))
```
nul800sebastiaan, line 63: You mean no member Id found.. copy/paste? 😉

OurUmbraco/Auth/AppBuilderExtensions.cs
```diff
@@ -0,0 +1,130 @@
...
+                            throw new InvalidOperationException("No project Id found in request"); // this will never happen
+
+                        //Get the stored auth key for this project and member
+                        var tokenService = new ProjectAuthKeyService(ApplicationContext.Current.DatabaseContext);
+                        var authKeys = tokenService.GetAllAuthKeysForProject(projectId);
```
nul800sebastiaan, line 68: Not entirely sure if we need to add member verification here as well? This method doesn't seem to validate the member Id is the owner of the project?

OurUmbraco/Auth/AppBuilderExtensions.cs
```diff
@@ -0,0 +1,130 @@
...
+                    }
+                }
+            });
+        }
+
+        private static bool EnsureHeaderValues(BaseContext context, params string[] headers)
+        {
+            foreach(var h in headers)
+            {
+                if (!TryGetHeaderValue(context, h, out _))
```
nul800sebastiaan, line 109: I never understood this notation, can we add a code comment or rewrite it as something a bit more easy to read?

OurUmbraco/Auth/ProjectAuthConstants.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace OurUmbraco.Auth
+{
+    public class ProjectAuthConstants
+    {
+        public const string ProjectIdClaim = "http://our.umbraco/projects/projectid";
+        public const string MemberIdClaim = "http://our.umbraco/projects/memberid";
+        public const string BearerTokenClaimType = "http://our.umbraco/projects";
```
nul800sebastiaan, line 7: These are not valid URLs? And not `https` - maybe the URL prefix should be configurable as well (so a global variable for the URL `https://our.umbraco.com/`. It has changes from `.org` to `.com` before and maybe in the future we'd want to move this to  `packages.umbraco.com` (I foresee we really might if we end up splitting up Our a bit more).

OurUmbraco/Project/Api/ProjectUploadController.cs
```diff
@@ -0,0 +1,111 @@
...
+        [HttpPost]
+        public async Task<HttpResponseMessage> UpdatePackage()
+        {
+            // Check if the request contains multipart/form-data.
+            if (!Request.Content.IsMimeMultipartContent())
+            {
+                throw new HttpResponseException(HttpStatusCode.UnsupportedMediaType);
+            }
+
+            string root = HttpContext.Current.Server.MapPath("~/App_Data");
```
nul800sebastiaan, line 41: This should be a different directory, at least a subdirectory under `~/App_Data` since it will pollute the root folder now and I don't see files being deleted afterwards.
Second note: we need to delete these temp files when we're done with them. :-)

OurUmbraco/Project/Api/ProjectUploadController.cs
```diff
@@ -0,0 +1,111 @@
...
+                    return Request.CreateErrorResponse(HttpStatusCode.InternalServerError, "No file attached");
+                }
+                else if (provider.FileData.Count > 1)
+                {
+                    return Request.CreateErrorResponse(HttpStatusCode.InternalServerError, "You can only upload one file");
+                }
+                else
+                {
+                    var packageFile = provider.FileData.FirstOrDefault();
+                    var fileName = packageFile.Headers.ContentDisposition.FileName.Replace("\"", "");
```
nul800sebastiaan, line 61: Hmm, need to make really sure that we can't do path traversal here.. if the filename ends in something like `../../web.config` maybe that would not be great? Not sure if this is a real vulnerability but need to look into that.

OurUmbraco.Site/Views/MacroPartials/Project/ApiKeys.cshtml
jmayntzhusen: Yes currently that is how it works, may extend to collabs later on 🙂

OurUmbraco/Auth/ProjectAuthConstants.cs
jmayntzhusen: Seems it just needs to be a unique string value..
https://docs.microsoft.com/en-us/dotnet/framework/wcf/extending/how-to-create-a-custom-claim

OurUmbraco/Project/Api/ProjectUploadController.cs
jmayntzhusen: As discussed on Slack, this seems to not be a problem anyways 🙂

OurUmbraco/Auth/AppBuilderExtensions.cs
jmayntzhusen: Added in https://github.com/umbraco/OurUmbraco/pull/552/commits/6de9aa156112210bd4d89f1e2f6d3450ec3cf6a6

OurUmbraco/Auth/AppBuilderExtensions.cs
jmayntzhusen: Fixed in https://github.com/umbraco/OurUmbraco/pull/552/commits/6de9aa156112210bd4d89f1e2f6d3450ec3cf6a6

OurUmbraco/Auth/AppBuilderExtensions.cs
jmayntzhusen: Added a note here: https://github.com/umbraco/OurUmbraco/pull/552/commits/2abb6b9c2ba681353753bbf957f05fd6041d6afb

OurUmbraco/Project/Api/ProjectUploadController.cs
jmayntzhusen: Fixed in https://github.com/umbraco/OurUmbraco/pull/552/commits/4e4bc287aafab1d7c967b67d66c6ecb5a2cf2c8b

nul800sebastiaan: Ah, I don't think you can delete the directory, if a few people upload files at the same time you'll get weird errors, but can just delete the temp file when the package has been created.

nul800sebastiaan: Fixed in 9124934a80be4034f1e147cc6a79b24d0fcd71a4

Issue #554, opened 2020-02-26
Unify Search on /packages with the one shown in the backoffice of umbraco


KevinJump:
At the moment the packages page on our is not using the same search as that used for the feed to the back office packages. we should unify them. 

in short we should make the packages page use the `PackageRepositoryService`

KevinJump:
Looking at this, it appears that the PackageRepositoryService returns packages which are retired. (e.g image processor in collaboration tools).

So just to confirm we should also fix that ?

jmayntzhusen:
Closing as the PR is merged, missed this issue at the time 🙂

PR #556, opened 2020-02-26
Unify Package search #554

KevinJump:
For #554 

- [x] Use the `PackageRepositoryService` to return the packages for our.umbraco.org. 
- [x] Fix the Service Search so it no longer includes retired packages
- [x] Add download sort order to the package Service 
- [x] Add download sort order option to the our.umbraco.org filters
- [x] don't reset the sort order filter when people pick a new umbraco version. 

this has extended the `PackageRepositoryService` a little, but we have not changed any of the existing information used by the back office. 

**Because the Service returns everything with the base URL when you are testing this, all the package results will have fully qualified URLs to our.umbraco.org. (not your local site).**

I am not sure how to check the back office elements are still working, can i change something in a standard umbraco install to point its package list to my own version of our ?

### DIFF
--- a/OurUmbraco.Site/Views/Partials/Projects/ListProjects.cshtml
+++ b/OurUmbraco.Site/Views/Partials/Projects/ListProjects.cshtml
@@ -1,7 +1,5 @@
-﻿@using Examine.SearchCriteria
-@using OurUmbraco.Our
-@using OurUmbraco.Our.Examine
-@using OurUmbraco.Project
+﻿@using OurUmbraco.Repository.Controllers
+@using OurUmbraco.Repository.Models
 @inherits OurUmbraco.Our.Models.OurUmbracoTemplatePage
 @{
     //This is the main Our project search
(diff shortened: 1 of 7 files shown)


KevinJump:
I've tested the back office (v8.5.3) when pointing to the updated search, the extra fields passed in the model have no effect on the packages screen inside Umbraco.

nul800sebastiaan:
Maybe a good idea to test with v7.15 as well. I expect it to be the same, but just in case.

KevinJump:
Hi, 

can confirm this also works against v7.15.3 

For reference I have done the following on both v8.5.3 and v7.15

### local version of our
- [x] copied the zip files for the starter kit and a few other packages into my local our repo.

### Umbraco Install
- [x] changed the `RestApiBaseUrl` to point to a local install of our
- [x] done a fresh install of Umbraco, including the starter kit
  - [x] confirmed Umbraco installs, and gets the starter kit from the local repo
- [x] browsed to packages inside Umbraco.
  - [x] confirm default listing is correct
  - [x] confirm viewing individual package pages still works
  - [x] test categories (e.g back office) still work
  - [x] test search works and returns list of packages
  - [x] install package from details page (assuming you have the zip file locally)

jmayntzhusen:
Awesome Kevin!

Will give this a final review Monday!

jmayntzhusen:
Also works for me, both the Our package listing page and a v7 and v8 sites 🎉

PR #558, opened 2020-03-05
New RSS feed URL

abjerner:
The RSS feed has a new URL on the new umbraco.com website 😮 

I noticed that `https://umbraco.com/blog/rss-uprofile-feed/` redirects to `https://umbraco.com/rss/blog/?tag=uprofile`.

Would a similar redirect be possible from `https://umbraco.com/blog/rss-feed/` to `https://umbraco.com/rss/blog/`? Then it doesn't break for people using the old URL.


hartviglarsen:
Fixed. 🙈

Thanks, Anders!

PR #560, opened 2020-03-16
Publish Packages with APIKey

Shazwazza:
This is a follow up to this PR https://github.com/umbraco/OurUmbraco/pull/552/files

I decided to target this PR to the other PR just so you can see the changes between that implementation and what I've done. If you want this implementation, you can merge this and then merge the other PR since it will now contain the changes in this one.

Changes:

* Re-updated references to some nuget packages
* No JWT since this is really unnecessary for what is being done here
* Authentication is done with HMAC Authentication which is a bearer token authentication similar to shared secret authentication but has an expiration, ensures that the token value is never the same between requests and ensures that it cannot be replayed (based on expiration). There's plenty of examples online about HMAC authentication and there are some nuget packages that offer this ootb but I've never tested them, instead this is taken and modified from Umbraco Deploy's implementation.
* Authentication and Authorization is done separately
  * Authentication is done at the OWIN layer using aspnets bearer token authentication handler. The identity assigned to the request contains the Claims that we want
  * Authorization is done with a custom attribute which simply checks that the required claims are in the request. Also means If authentication is ever changed, enhanced or more authn methods are added, all we require is the correct claims to be added to the identity
* Have changed a bunch of naming conventions. Most of this authn/authz stuff is specifically to do with the member packages but it was called something like "UmbracoToken" which isn't quite right so the names of all of this have been updated to reflect that. References to 'Token' in most cases have been changed to 'Key' because we store an API Key in the DB, not a Token. A Token is what is put into the request.

I've added a commented out example of how you might go creating an authenticated request via webapi. Depending on the CLI client that is being built, its possible that this method and code can just be used. Much of the HMACAuthentication class will need to be copied to the client along with some extension methods. If the CLI client is not written in c# (not sure?) it's still absolutely possible to create this request. I've created something in powershell that does this already but it could be done with any other language too.


jmayntzhusen:
Will test this with the CLI tool later today, thanks @Shazwazza 🙂

Quick note for myself - we need to set the new owin startup in the vsts web config as well as it is not source controlled but is the one we use in Devops.

jmayntzhusen:
@Shazwazza 

I made a few changes, only notable ones are that I moved the new owin startup settings into the existing one used on Our.
Additionally made some tiny changes to the claims verification in https://github.com/umbraco/OurUmbraco/pull/560/commits/2e740af2bf046eb61b20ed0be782675ecbd5d38b

As the `ClaimsIdentity.GetUserId` method automatically uses a different Claims type I changed it to `FindFirstValue` with our own specified Claims type.

For a full overview of the changes you can see this compare: https://github.com/umbraco/OurUmbraco/compare/4f94779d940b6e593d032d3e704e764857d954f6...afd7dd53a5dc9c0d084803d231aa011f894532a1

One question about this implementation - is the API key still something to be showed once only to the user? Or can we show it on each visit?

TODO's:
- Update UI dependent on whether we show the token once or each time
- Add support for generating a new key and overwriting existing (UI mostly the methods are there)

If you'd like a look at how a working example of the CLI client has the request set up it is in this PR: https://github.com/umbraco/Umbraco-Packager-CLI/pull/7

Shazwazza:
@jmayntzhusen looks good, i added one comment about a null check.

> One question about this implementation - is the API key still something to be showed once only to the user? Or can we show it on each visit?

This should by default be hidden, just like any apikey. This is still the 'secret' key that builds the token, this is the 'secret' key used in the HMAC hashing. If this is exposed to others then they can re-create the hash. There's 2 diferent approaches to take: Show only one time, else you need to generate a new one, or just not show the value unless the user clicks specifically on show value. I suppose the latter is slightly less secure but i don't think it's much of an issue, i see that option being used in many services, even things like Stripe.

jmayntzhusen:
Will merge this in for now as it has been verified by Shannon. That means it will be merged into https://github.com/umbraco/OurUmbraco/pull/552/files not into master 🙂

I have some additional changes that will be on top for a final review soon.

PR #561, opened 2020-04-03
Managing api keys

jmayntzhusen:
This is an extension to https://github.com/umbraco/OurUmbraco/pull/552/files that handles the frontend parts.

Has a migration to create a new node with a macro to load a new partial for managing api keys. As well as the view itself and a new API controller to manage form submissions, etc and talk to the key service.

Here is a gif of the workflow:
apiKey


jmayntzhusen:
Made some changes based on feedback during todays hackathon!

1. Now allows more than 1 key per person
2. Changes the key overview to be a bit smaller and table based instead of huge cards that are unmanageable with a lot of them
3. Changed the key format to be "projectid-memberid-token" so something like this: 
9213-4576-AC58BD65A90F132C5AE362F42E24B5B69EA717D74068D8074EC79BA6C7FC47FA

Means we can "just" do one thing to copy for the CLI tool:

pkgtooltest

PR #563, opened 2020-04-11
Reorganising member package listing to be under 'status' headings

LottePitcher:
This relates to https://github.com/umbraco/Umbraco.Packages/issues/56

Previously all packages were listed in alphabetical order with no indication of their 'status', i.e. whether they were Active, Draft or Retired

This PR now groups members' packages by status, with Active, then Draft, then their Contributions (already a separate listing), followed by 'Retired'. Hopefully this will encourage package developers to retire active packages more often.

For example:

image

image

image

image

If they have no packages then no headings appear, just the 'add' button:

image

### DIFF
--- a/OurUmbraco.Client/src/js/app.js
+++ b/OurUmbraco.Client/src/js/app.js
@@ -37,9 +37,10 @@ jQuery(document).ready(function () {
     // Tab
     $('.tabs li').click(function () {
         var tab_id = $(this).attr('data-tab');
+        var tab_container = $(this).closest("div");
 
-        $('.tabs li').removeClass('current');
-        $('.tab-content').removeClass('current');
+        $(tab_container).find('.tabs li').removeClass('current');
+        $(tab_container).find('.tab-content').removeClass('current');
 
         $(this).addClass('current');
         $("#" + tab_id).addClass('current');
--- a/OurUmbraco.Client/src/scss/elements/_profile.scss
+++ b/OurUmbraco.Client/src/scss/elements/_profile.scss
@@ -283,8 +283,8 @@
 
 .profile-settings-packages {
 	.button {
-		margin-top: 30px;
-		margin-bottom: 30px;
+		margin-top: 10px;
+		margin-bottom: 10px;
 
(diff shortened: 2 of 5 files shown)


LottePitcher:
Am now wondering whether the status of the project should be addressed in the 'Contributors' listing.  Just investigating...

jmayntzhusen:
How about doing something like this:

```
- Personal packages
-- Active
-- Draft
-- Retired

- Contributions
-- Active
-- Draft
-- Retired
```

Then through a bit of JS we could make all of these elements collapsible? Maybe default retired to collapsed fx.

LottePitcher:
Plan is now to have two separate headings, one for own packages and one for contributions. Then to have tabs under each heading to navigate between each status (with counters).  So something like

My Packages
|  Active (3)  |  Draft (0)  |  Retired (3)  |

each tab content can have intro text confirming what the different status means

LottePitcher:
This PR now changes the package listing to show two separate sets of tabs, one for My Packages and one for My Package Contributions:

image

image

The tabs are implemented in the same way as the tabs at the bottom of the package details page (for package files, archived files etc).  app.js has been updated because that JavaScript assumed that there was only one set of tabs on a page, so I needed to tweak that code to ensure that it was scoped to an outer container div.  I tested that the tabs on package details still worked and, as far as I could tell, that is the only other place that uses this 'tabs' feature.

I don't know if anyone has any better ideas for the styling of the tab content.  The package details page uses a background of grey for the active tab but that seemed OTT here as the package boxes definitely look best on a white background. As you can see in the screenshots I've just reversed the logic so the inactive tabs are background grey.

LottePitcher:
This PR should also be checked for a member who has no packages.  Would it be better to hide all the this and display nothing beneath the Add Package button?

jmayntzhusen:
Works perfectly when you have packages:
image


When you don't have any packages it looks like this:
image


I'd say we could probably improve the "empty" list a bit, but it's still better than what we have now, so let's get this merged in and we can consider future improvements later 🙂

PR #564, opened 2020-04-11
updating logic for showing Karma on a package to work when no karma yet

crgrieve:
Fix for issue here:
https://github.com/umbraco/Umbraco.Packages/issues/31

Previously showing -1 when no Karma. Now showing 0 as expected, see screenshot:
image


nul800sebastiaan:
SuperTak!

nul800sebastiaan:
Just as an explanation for "future me":  the `GetInt` method ends up returning a `-1` if the SQL query doesn't return any results. You'd expect it to be the default value 0 but this is a special case to indicate a `null` result. We'd never want to return `-1` for package votes anyway, so this fix is great.

PR #565, opened 2020-04-11
Packages version list

crgrieve:
As per packages issue: https://github.com/umbraco/Umbraco.Packages/issues/55

Before:
image

After:
image


LottePitcher:
Looks good to me @crgrieve - will mean those package developers with long version history will be able to reach those action buttons without having to use the browser 'zoom out' function!!  Worth getting deployed soon I reckon ;-)

jmayntzhusen:
Hadn't made the connection that it was one long line because it was essentially 1 "word" 😮 nice simple change - thanks a lot @crgrieve ! 🎉

Issue #568, opened 2020-04-17
Added blog not showing up on Community blog-posts


jvtroyen:
Almost exactly 1 year ago, I added my blog to CommunityBlogs.json.
It took me until now to add more topics (I know, I know, I was born lazy). I see they are not automagically appearing in the list.
The RSS Url I added was http://jvantroyen.blogspot.com/feeds/posts/default, in which I can see the new posts.
Is there another step involved? Did I miss something? Am I looking in the wrong place?

abjerner:
Hi @jvtroyen,

Your feed seems to be an ATOM feed, which is not supported. ATOM feeds have a `<feed>` element at the root, and each post is expressed by an `<entry>` item.

RSS feeds have a `<rss>` element at the root, and then consists of a single `<channel>`, which again contains an `<item>` element for each post.

jvtroyen:
Hi @abjerner ,
that is very good to know. I never used RSS before, so wasn't aware of this difference. I'll try to see what I can do about this format.

jvtroyen:
The Blogger feed is ATOM, but you can convert it using FeedBurner.
I put the new url (https://feeds.feedburner.com/MyUmbracoAdventures) in a PR.

Thanks @abjerner !

PR #569, opened 2020-04-20
Converted ATOM feed to RSS feed

jvtroyen:
See #568. I added my feed last year, but turns out it's an ATOM feed, which is not supported. 
Now converted to RSS using FeedBurner.


nul800sebastiaan:
Thanks @jvtroyen - that works a lot better! 👍 
image

jvtroyen:
Thank you @nul800sebastiaan

PR #571, opened 2020-04-21
Update to Documentation ordering

OwainWilliams:
Adding Heartcore and ordering the nav in a more logical order

### DIFF
--- a/OurUmbraco/Documentation/DocumentationUpdater.cs
+++ b/OurUmbraco/Documentation/DocumentationUpdater.cs
@@ -223,6 +223,15 @@ namespace OurUmbraco.Documentation
                         case "frequently-asked-questions":
                             return 6;
                         
+                        //Umbraco Heartcore
+                        case "getting-started-cloud"
+                            return 0;
+                        case "api-documenation"
+                            return 1;
+                        case "client-libraries"
+                            return 2;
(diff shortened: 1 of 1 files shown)


OurUmbraco/Documentation/DocumentationUpdater.cs
```diff
@@ -223,6 +223,15 @@ private SiteMapItem GetFolderStructure(DirectoryInfo dir, string rootPath, int l
...
                         
+                        //Umbraco Heartcore
+                        case "getting-started-cloud"
+                            return 0;
+                        case "api-documenation"
+                            return 1;
+                        case "client-libraries"
+                            return 2;
+                        case "versions-and-updates"
+                            return 3;
```
sofietoft, line 234: Looks great @OwainWilliams !
I've made a few suggestions: looks like a colon needs to be added after each case.

```suggestion
                        case "getting-started-cloud":
                            return 0;
                        case "api-documenation":
                            return 1;
                        case "client-libraries":
                            return 2;
                        case "versions-and-updates":
                            return 3;
```

jmayntzhusen:
Looks good 🙂🎉

Will merge in now, will be on the site after the next deployment (can't say how soon that is atm 😋)

PR #573, opened 2020-04-21
Improved the look of the buttons on the package list

Rockerby:
A quick refresh of the edit / forum / team buttons on the package list to better go with the rest of the site.

Desktop:

image

Mobile:

image

### DIFF
--- a/OurUmbraco.Client/src/scss/elements/_profile.scss
+++ b/OurUmbraco.Client/src/scss/elements/_profile.scss
@@ -427,6 +427,39 @@
 			margin-top: 0 !important;
 		}
 	}
+
+	.list-buttons {
+		display: inline-block;
+		margin-top: 10px;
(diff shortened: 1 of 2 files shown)


jmayntzhusen:
Thanks @Rockerby 

This is definitely an improvement! Just a thought - any reason you chose this lightblue and not the one the other buttons have, like this?

image


Not sure we have any ui with sort of "secondary" buttons with a lighter blue currently, although it could make sense..

Rockerby:
Thanks Jesper,

I made them light blue to copy from the more 'inline' button styling of the forums found for categories. I'm not precious if you think they'd look better in dark blue :)

jmayntzhusen:
Makes sense, forgot we do have this other blue color in the forum and frontpage. Will merge - thanks again! 😁

PR #576, opened 2020-04-30
catch exceptions while indexing yaml meta tags

dampee:
If we don't catch errors and there is a syntax error in the meta tags, indexing will stop and search will not be working again (for documentation) on our.

### DIFF
--- a/OurUmbraco.Site/App_Plugins/OurManagement/ourmanagement.controller.js
+++ b/OurUmbraco.Site/App_Plugins/OurManagement/ourmanagement.controller.js
@@ -142,4 +142,21 @@
                 notificationsService.error(vm.memberError);
             });
     };
+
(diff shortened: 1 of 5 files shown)


jmayntzhusen:
Looks good, I especially like that we can now search for +tags: yamlissue in the index to find the error articles!

Would like to have it appear as some kind of warning somewhere so we don't have to remember to do it ourselves 🤔 May submit another PR to add the results of that to the backoffice dashboard that Sofie also has access to..

dampee:
Maybe add to my branch, so it is merged in together ?

jmayntzhusen:
Made a few additions, there is a now a new backoffice button for @sofietoft to use:

image

When she clicks it it will give a list of files with invalid YAML (if any) 🙂

dampee:
This is an awesome simple addition @jmayntzhusen

PR #590, opened 2020-06-22
Grabs pkg version from form and updates the content node accordingly

jmayntzhusen:
This is conjunction with https://github.com/umbraco/UmbPack/pull/27 will ensure that a new package pushed via UmbPack will update the current version to whatever the package file has set in its package.xml:

pkgVersion


LottePitcher:
Looks good to me @jmayntzhusen, will be good to see this fixed! 🎉

LottePitcher:
Although having read latest on umbraco/UmbPack#27 I can see that you'll probably need to change it so it only updates the version property if the form data has an item with the 'packageVersion' key. As I figure it's easiest to let umbPack decide whether to send it over or not, and Our just updates it if it's provided?

PR #597, opened 2020-06-30
Add archive endpoint for UmbPack

jmayntzhusen:
This will work with https://github.com/umbraco/UmbPack/pull/28 to enable UmbPack to archive specific package files.

### DIFF
--- a/OurUmbraco/Project/Api/ProjectUploadController.cs
+++ b/OurUmbraco/Project/Api/ProjectUploadController.cs
@@ -26,6 +26,17 @@ namespace OurUmbraco.Project.Api
             return files.Where(x => x.FileType == "package" || x.FileType == "hotfix").OrderByDescending(x => x.Version.Version).ToList();
         }
 
+        [HttpGet]
+        public int GetCurrentPackageFileId()
+        {
+            var contentService = ApplicationContext.Services.ContentService;
+            var packageEntity = contentService.GetById(ProjectNodeId);
+            
+            var packageFileId = packageEntity.GetValue<int>("file");
+            
+            return packageFileId;
+        }
+
         /// <summary>
         /// Endpoint that takes a list of package file ids from UmbPack then archives them
         /// </summary>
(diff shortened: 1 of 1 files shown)


OurUmbraco/Project/Api/ProjectUploadController.cs
```diff
@@ -24,6 +26,37 @@ public IEnumerable<WikiFile> GetProjectFiles()
...
+        public HttpResponseMessage ArchiveProjectFiles(IEnumerable<int> ids)
+        {
+            var mediaProvider = new MediaProvider();
+
+            try
+            {
+                foreach (var archiveFileId in ids)
+                {
+                    var archiveFile = mediaProvider.GetFileById(archiveFileId);
+                    archiveFile.Archived = !archiveFile.Archived;
```
mattbrailsford, line 44: Should `Archived` just be set to `true` rather than it toggling it's current value as it would be kind of weird behavior calling ArchiveProjectFiles and it un-archived a file?

jmayntzhusen: Good catch 🙂 Too quick to copy paste code 😅

mattbrailsford:
How come this code isn't present in the main branch? Is this why the UmbPack archive flag isn't working because this endpoint has gone missing? https://github.com/umbraco/UmbPack/issues/40

nul800sebastiaan:
@mattbrailsford super strange! I was going to say maybe it has something to do with the rename from `master` to `main` but that was weeks later.. I don't see it in the history of this file though! 🤔 

Let's see what I can do!

mattbrailsford:
Yea, branch rename was gonna by my primary suspect, but very strange.

nul800sebastiaan:
@mattbrailsford Try again, the code has been deployed now!

mattbrailsford:
@nul800sebastiaan looks like that did the trick 👍

nul800sebastiaan:
Woohoo! 🎉

PR #600, opened 2020-07-15
updating backoffice username

LottePitcher:
Changed readme with the correct backoffice username from new DB.  

However... I would find local development easier if both the backoffice and the member login had the same format.  So perhaps the username could be changed instead of accepting this PR?!


jmayntzhusen:
New db will be up momentarily with more consistent login info - thanks for spotting that! 😉

PR #601, opened 2020-07-15
fixes required to rebuild 'projectIndexer' with new dev db

LottePitcher:
Using the new dev db, rebuilding 'projectIndexer' returned 0 documents for me because of two errors.

### Error 1
As there is no data in the ```projectDownload``` table, the most recent download date returned is ```DateTime.MinValue```.  The code then falls over when trying to subtract 365 days from the returned date (of course can't have a date earlier than MinValue). This will only be an issue when using the new db. This PR changes it to not query the db and just return empty collection in this instance.

### Error 2
The other error was that the database was returning NULL for the ```removedBy``` column for the one record in the ```wikiFiles``` table, and the ```RemovedBy``` property of the ```WikiFile``` class is a *non-nullable* int.  I've implemented a workaround so am now able to get the index rebuilt and can see the 2 packages on the front-end 🎉.  Not obvious to me why this would only cause a problem with this new db as surely lots of live ```wikiFiles``` records would have null ```removedBy``` properties...


jmayntzhusen:
Thanks a lot @LottePitcher this works perfectly 🎉

PR #603, opened 2020-07-22
Calculates the nuget downloads for 6 months (not days) as per comment

KevinJump:
Comment in code says `// add the nuget downloads for the last 6 months` but actually it was only doing it for the last 6 days. 

updated to work it out for last 6 months. *(also want to confirm my re naming of branches works)*

### DIFF
--- a/OurUmbraco/Our/Examine/ProjectPopularityPoints.cs
+++ b/OurUmbraco/Our/Examine/ProjectPopularityPoints.cs
@@ -105,7 +105,7 @@ namespace OurUmbraco.Our.Examine
 
             if (_dailyNugetDownLoads.HasValue)
             {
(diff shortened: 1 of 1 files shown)


KevinJump:
It might be that this dis-proportionally skews scores in favor of Nuget downloads, at the higher end this is is adding approx 30k-50k to the 'score' value for a package, (at the moment the highest scoring package is approx 45k). 

so might have always been the intention to weight the nuget scores lower and the comment is just wrong ?

jmayntzhusen:
Testet this out locally with a 3 months old live db copy. The order of the frontpage of "popular" packages didn't change whatsoever. I see the popularity score of fx Umbraco Forms went from ~17k to ~20k after pulling down the updated nuget downloads.

I think we can merge this, will not change much but be more correct and give a slight boost to packages with tons of nuget downloads (as it should 🙂).

PR #607, opened 2020-08-13
Adding lifetime MVP badge to profiles

callumbwhyte:
In the midst of @leekelleher [moaning](https://twitter.com/leekelleher/status/1293900765052850181) about 10 awards not being good enough - _kidding, of course_  😜 - it was discovered that there's no "lifetime MVP" badge on Our.

Thought I'd add one!

I assumed I'd have to create a migration for introducing the new role but it seems migrations were [mysteriously removed](https://github.com/umbraco/OurUmbraco/commit/0fd7b25)... Happy to add one if someone can tell me where!


abjerner:
@callumbwhyte what a great way to show that @leekelleher is not a lifetime MVP, but only a 10x MVP 😆 

The migration file seems to be back here: https://github.com/umbraco/OurUmbraco/blob/main/OurUmbraco/Our/MigrationsHandler.cs

I think the old migration file would primarily make changes to the database (eg. adding content, properties etc.), but as a new database is now available, the old file was canned. @nul800sebastiaan can probably confirm to be sure.

callumbwhyte:
I found the migration file, but as it was completely emptied a couple of months back I wasn't sure what to do... Happy to just add one there if that's the done thing.

I knew there was a new database being created but didn't know it was ready yet? As such this change is totally untested... Sounds very mergeable, right? 😅

LottePitcher:
Well if you do want to test it first... a working link to download the DB is on https://github.com/umbraco/ourumbraco!

nul800sebastiaan:
This looks cool @callumbwhyte! Love the infinite symbol there! 😉

If you can add the migration then that would be super nice! You can look at the history of that file for inspiration.

nul800sebastiaan:
The lifetimers should also be at the top of this list: https://our.umbraco.com/community/most-valuable-people/ - could you have a look at adding them there as well please? 🙏

nul800sebastiaan:
We'll re-arrange the list later. I've updated some of this to make it look a bit nicer but other than that all good, thanks @callumbwhyte!

PR #620, opened 2020-09-07
Fix #619 - Nuget Package download scores.

### LINKED ISSUE #619: Nuget scores are a bit off, if previous nuget packages are unpublished.
It looks like there are some cases where new packages will be given higher scores than they should because of the way the nuget package is published/managed. 

image

When we work out the nuget score we attempt to get the published date and then work out the average by dividing total downloads / published date.

but if the developer unpublished all the old versions of the package then the publish date becomes the last one that is 'live' - this means we can get the wrong date (e.g today) and end up with a massive total per day - because the total is all downloads including now unpublished packages.

- solution is to use the `commitDate` of a package in nuget not the `publisheddate` (so the day it is uploaded to nuget) - in most cases this will be the same date. 

- also when we calculate the score we do average * 6 months - if the package is younger than 6 months we should do avarage * package age.

*pr incoming*

KevinJump:
Fix for the nuget scores 

 - tidied up NugetPackageDownload Service 
 - use the CommitDate not the PublishedDate
 - calculate 6 month downloads for nuget from either 6 months or package create date (whichever is smaller). 

(will need testing with the real data)

### DIFF
--- a/OurUmbraco/Community/Nuget/NugetPackageDownloadService.cs
+++ b/OurUmbraco/Community/Nuget/NugetPackageDownloadService.cs
@@ -43,124 +43,188 @@
 
             if (projects.Any())
             {
-                // get the services from nuget service index
-                var restClient = new RestClient(this._nugetServiceUrl);
-
-                var result = restClient.Execute(new RestRequest());
-
-                var response = JsonConvert.DeserializeObject<NugetServiceIndexResponse>(result.Content);
-
-                if (response != null)
+                var searchUrl = GetNugetService("SearchQueryService");
+                if (!string.IsNullOrWhiteSpace(searchUrl))
                 {
-                    // get a url for the search service
-                    var searchUrl = response.Resources.FirstOrDefault(x => x.Type == "SearchQueryService")?.Id;
+                    var nugetPackageDownloads = new List<NugetPackageInfo>();
 
-                    if (!string.IsNullOrWhiteSpace(searchUrl))
+                    // we will loop trough our projects in groups of 5 so we can query for multiple packages at ones
+                    // the nuget api has a rate limit, so to avoid hitting that we query multiple packages at once
+
+                    foreach (var projectGroup in projects.InGroupsOf(5))
                     {
-                        var nugetPackageDownloads = new List<NugetPackageInfo>();
+                        var packageQuery = GetNugetPackageQuery(projectGroup);
 
-                        // we will loop trough our projects in groups of 5 so we can query for multiple packages at ones
-                        // the nuget api has a rate limit, so to avoid hitting that we query multiple packages at once
-
-                        foreach (var projectGroup in projects.InGroupsOf(5))
+                        if (!string.IsNullOrWhiteSpace(packageQuery))
                         {
-                            var packageQuery = string.Empty;
+                            var searchQuery = $"{searchUrl}?q={packageQuery.TrimEnd("+")}&prerelease=true";
+                            var packageSearchResult = GetNugetSearchResults(searchQuery);
 
-                            foreach (var project in projectGroup)
+                            if (packageSearchResult != null)
                             {
-                                var nuGetPackageCmd = GetNuGetPackageId(project);
-
-                                if (!string.IsNullOrWhiteSpace(nuGetPackageCmd))
+                                foreach (var package in packageSearchResult.Results)
                                 {
-                                    packageQuery += $"packageid:{nuGetPackageCmd}+";
-                                }
-                            }
-
-                            if (!string.IsNullOrWhiteSpace(packageQuery))
-                            {
-                                var searchQuery = $"{searchUrl}?q={packageQuery.TrimEnd("+")}&prerelease=true";
-
-                                restClient = new RestClient(searchQuery);
-
-                                var packageResponse = restClient.Execute(new RestRequest());
-
-                                var packageSearchResult =
-                                    JsonConvert.DeserializeObject<NugetSearchResponse>(packageResponse.Content);
-
-                                if (packageSearchResult != null)
-                                {
-                                    foreach (var package in packageSearchResult.Results)
+                                    var packageInfo = GetNugetPackageInfo(package);
+                                    if (packageInfo != null)
                                     {
-                                        var packageInfo = new NugetPackageInfo
-                                                              {
-                                                                  PackageId = package.Id,
-                                                                  TotalDownLoads = package.TotalDownloads
-                                                              };
-
-                                       
-
-                                        // try get details about downloads over time
-                                        // so we get the publish date of the package on nuget. And calculate the average downloads per day.
-                                        // we can use this data for the popular package query on our
-                                        // we can run into issues when a package has more than 128 versions. This call will return a different response then
-                                        // see https://docs.microsoft.com/en-us/nuget/api/registration-base-url-resource
-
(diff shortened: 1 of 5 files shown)


dawoe:
Looks good to me. I did not do a test with real data.

jmayntzhusen:
Testing with a fresh copy of the live db I get weird results after doing a Nuget download import and rebuild of the project index from this pr.. Here is my local test on the left, and live on the right sorted by popularity. The db copy is from last week so download counts are almost the same..
image

Sorry about how long this has taken to get approved, but seems more testing is still required, haven't been able to find why the results are being strange here yet.

dawoe:
Hi @jmayntzhusen 

Could it be that you the nuget import did not succeed ? Because none of the packages with a lot of nuget downloads is visible on your local copy.

Could you check uSync for instance on your local copy and see how may nuget downloads that has ?

Dave

jmayntzhusen:
Yep had the same thought, here is local usync vs live:
image

Same for Forms, etc. Need to step through the popularity points, but ran out of time for it for today.

KevinJump:
looking at the responses I think the commitDate can't be trusted. 😞 

Part of this fix was to push past the publishdate, because if someone unpublishes all their old versions of a package, then the package looks new, and shoots up (because its downloads per day go massive). 

On a recent package commitDate is fine, but on older projects it is likely to have been reset. - I will look at the code see if we can use PublishDate and still workout when something isn't as new as it seems. 

that however isn't 100% the reason for the points issue, but it might be related.

KevinJump:
Hopefully sorted. 

1. where were trying to calculate the value based on the _now value, and the _now value isn't now, its passed in and is actually last modification date of the package on our. 
2. where were doing the date thing wrong i'd missed a - of the add months so it was returning a negative value (so giving packages a negative score!)

also. 
The commit date from nuget isn't reliable, I suspect its more of an internal thing as lots of my results came back with the same date in Feburary. 

To mitigate the oldest date is now calculated as the oldest of publishedDate and commitdate (when the publishedDate is not 1900).

jmayntzhusen:
Ran this again locally with all the packages my local db from October knows about, here are the changes after a fresh Nuget import (the Our download score wouldn't be the same as Oct 22nd when the backup was taken):
Left is the local version based on this PR. Right is current Our as of taking the screenshot
image

A bit concerned that fx DiploGodmode is gone while Contour is suddenly back 🤔 This popularity score stuff is quite hard to get right..

KevinJump:
there is a file stored in ~/App_Data/TEMP/NugetDownloads that should give you an indication as to weather the nuget download bit has worked.  that should have total and avarage downloads per day for each package. 

so it might be worth comparing a few to Nuget to see if that match ?

KevinJump:
Its the oldest date bit from nuget which is hard. 
https://github.com/umbraco/OurUmbraco/blob/551f289580b4b3457c295cd0c6d37c2a4c202a3f/OurUmbraco/Community/Nuget/NugetPackageDownloadService.cs#L161

Sometimes the published date is right, then it looks like the comit time stamp is more reliable, then they reset it :(

KevinJump:
image

not sure that was me ???

jmayntzhusen:
Hmm so fx the blob storage package has this:
```json
    {
        "PackageId": "UmbracoFileSystemProviders.Azure",
        "TotalDownLoads": 638995,
        "AverageDownloadPerDay": 345
    },
```

Where fx diplo godmode has
```json
    {
        "PackageId": "Diplo.GodMode",
        "TotalDownLoads": 84990,
        "AverageDownloadPerDay": 2
    },
```

Thing is.. most packages have 1-4 with a few much higher ones. Maybe it is just the weighting of that score that needs to be reduced? Seems the effect we are seeing is that a few with a ton of daily downloads gets bumped up a lot, while most others are unaffected or even bumped down below the ones that aren't listed on NuGet.

KevinJump:
Hum, so diplo.godmode should be getting closer to 51 a day (it's nuget score). 

I might put it in locally to see how we are working out 2. :(

KevinJump:
Humm.. when i run it locally with Dilplo.GodMode as a package it gets 52 downloads a day. which is (close enough) to the nuget number of 51.

KevinJump:
Hi, 

I've added the PackageDate to the json, (not used anywhere but it means we can see it). 

the downloads per day - is calculated as days since the package date / total downloads. so we should see from this if we can get to the bottom of it only showing 2 for you. 

for reference this is the URL that is used to get the json with all the dates in it. 
https://api.nuget.org/v3/registration5-semver1/diplo.godmode/index.json

jmayntzhusen:
Great, thanks Kevin - I'll run it again 🙂

And just so we are on the same page, the following should be what is required to update it, right?:
- Run Hangfire job `NugetPackageDownloadService.ImportNugetPackageDownloads`
- Reindex projectindex
- Visit package page and check the ~/App_Data/TEMP/NugetDownloads file?

KevinJump:
yep, think so, although the package file in app_data gets written by the first step, so you don't need to reindex to read that.

but the index updates the scores and hence the package listings on the site.

jmayntzhusen:
Hmm tried stepping through this:
https://github.com/umbraco/OurUmbraco/blob/551f289580b4b3457c295cd0c6d37c2a4c202a3f/OurUmbraco/Community/Nuget/NugetPackageDownloadService.cs#L161

And the date var is being set to the "blankdate" even though the items have published dates and timestamps:
image

Wondering if the orderby on line 169 may differ based on my computer culture or something 😕

KevinJump:
maybe commit date is diffrent on diffrent nuget servers ? _ because that can be the only place with a null date? 

maybe . 

change x.commitTimeStamp to ?

```
x.CommitTimeStamp > blankDate ? x.CommitTimeStamp : DateTime.Now, 
```

KevinJump:
nuget is using an UTC timestamp format. maybe we need to be explicit when serializing it?

jmayntzhusen:
The commitTimeStamp and CatalogEntry.PublishedDate are both set though, and higher values than year 1900, you can see their values in the bottom right panel of my screenshot above. Even still it set the date to the blankDate in the big linq query 😕

KevinJump:
I am experimenting with `[JsonConverter(typeof(IsoDateTimeConverter))]` on the dates i will push something if that seems to work.

it should overcome server settings (not sure they would matter here, but you never know)

jmayntzhusen:
Broke up the linq statement into smaller bits, and the ordering is indeed the problem. It orders them exactly right, but in ascending order so when we run FirstOrDefault afterwards we get the lowest value which was the blankDate one:
image

A bit surprised it works differently on your end tbh?

KevinJump:
Another lesson in the "big linq statements are not clever" column 🤔

jmayntzhusen:
Still getting the same problem, and I found that it's because NuGet seems to be returning the value from their API.. For this call:
https://api.nuget.org/v3/registration5-semver1/diplo.godmode/index.json

I get 
image

So it has the published data as `1900-01-01T00:00:00+00:00` which is what we get returned from the GetOldestDateFromEntries method.

KevinJump:
I think this PR is dead ☠️ - the other dates nuget bring back are not consistant. so it doesn't look like you can do this to get a date other than publish date . 

besides the other pr #753 superseeds this and puts in a simpler fix for large packages so hopefully can be included.

PR #631, opened 2020-09-29
Grant member karma

jmayntzhusen:
This PR will make it possible for me to add enough karma to new users to be able to upload packages.
It uses the existing backoffice dashboard, here is an example 🙂
grantKarma

Main question I have is whether we should set both the total and current rep points? Not sure what the difference between them is, all I know is the total rep points need to be above 30 for the package upload part.

### DIFF
--- a/OurUmbraco.Site/App_Plugins/OurManagement/ourmanagement.controller.js
+++ b/OurUmbraco.Site/App_Plugins/OurManagement/ourmanagement.controller.js
@@ -143,6 +143,23 @@
             });
     };
 
+    $scope.addKarmaToMember = function (memberId) {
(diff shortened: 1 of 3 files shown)


abjerner:
I find it hard to believe @nul800sebastiaan only has <s>30</s> 31 karma points 😆

nul800sebastiaan:
Haha!

Well, I set it to 71 for both when requested. This is because the limit for the spam filter in the forum is 70. Would be a bit sad if they can't even comment in their support forum! 😅

jmayntzhusen:
Good point @nul800sebastiaan 🙂

https://github.com/umbraco/OurUmbraco/pull/631/commits/4859639ae2de57e8a311ef54bf543ccc3e6e57f4 sets both Karma values to 71 now, and it #worksonmymachine..

nul800sebastiaan:
Works for me too! Deploying now. 👍

PR #634, opened 2020-10-03
633 Link to packages from the profile window

AaronSadlerUK:
Added a packages link to the quick menu

Fixes #635 
Fixes https://github.com/umbraco/Umbraco.Packages/issues/65

### DIFF
--- a/OurUmbraco.Site/Views/Partials/Members/QuickMenu.cshtml
+++ b/OurUmbraco.Site/Views/Partials/Members/QuickMenu.cshtml
@@ -67,9 +67,9 @@
                         <div class="public-profile">
                             <a href="@Url.GetProfileUrl(profile)">View my public profile</a>
                         </div>
-                            <div class="public-profile">
-                                <a href="/member/profile/packages/">View my packages</a>
-                            </div>
+                        <div class="public-profile">
+                            <a href="/member/profile/packages/">View my packages</a>
+                        </div>
                         @if (profile.IsAdmin)
                         {
                             <div class="public-profile">
--- a/OurUmbraco.Site/Views/Partials/Members/QuickMenu.cshtml
+++ b/OurUmbraco.Site/Views/Partials/Members/QuickMenu.cshtml
@@ -5,14 +5,6 @@
 
 @{
     var profile = MemberData;
-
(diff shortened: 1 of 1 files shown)


OurUmbraco.Site/Views/Partials/Members/QuickMenu.cshtml
```diff
@@ -2,8 +2,17 @@
 @using OurUmbraco.Forum.Extensions
 @using OurUmbraco.Our.Extensions
 @using OurUmbraco.Our.Models
+
 @{
     var profile = MemberData;
+
+    var homePage = Umbraco.TypedContentAtRoot().FirstOrDefault();
+    var profilePage = homePage.Descendants("Textpage").Where(x => x.Name == "Profile").ToList().FirstOrDefault();
```
nul800sebastiaan, line 10: `Descendants` is very expensive and this needs to be done for each logged in user each time. 

We're very happy just hardcoding this to `/member/profile/packages/` for now.

nul800sebastiaan (changes requested):
It's not advise I give often, but can you please change it to a hardcoded route? 👍

OurUmbraco.Site/Views/Partials/Members/QuickMenu.cshtml
abjerner: @nul800sebastiaan wouldn't it make sense to introduce a constant here?

https://github.com/umbraco/OurUmbraco/blob/main/OurUmbraco/Constants.cs

Either for the route directly, or the key/GUID of the packages page?

nul800sebastiaan: No key/GUID please. Can't guarantee the dev db has the same GUID.

I don't really think we need to write code to put a string in a single place for now. When the packages section gets overhauled it might be something we need.

AaronSadlerUK:
Sure ill update that shortly when I get back home

OurUmbraco.Site/Views/Partials/Members/QuickMenu.cshtml
AaronSadlerUK: I have made that change 👍

nul800sebastiaan:
Thank you @AaronSadlerUK  - that works! Will release soon.

PR #635, opened 2020-10-07
Track downloads from main download-page

filipbech:
We already do this on the individual release-pages. Now we should get more real data. 

This is done blindly, as I'm on my mac today - so please test :-p @nul800sebastiaan


filipbech:
In order for it to be consistent with the others, we should probably use .Name instead of .FullVersion

nul800sebastiaan:
@filipbech Good call! If only.. we had `.Name` available. Fake it 'til you make it.. https://github.com/umbraco/OurUmbraco/pull/635/commits/634d9f2082f30ee08eb88e4858944ceb77c27225

PR #639, opened 2020-10-16
452 Add breadcrumb to the search result entries

### LINKED ISSUE #452: Add breadcrumb to the search result entries
When searching for documentation it would be nice to include the breadcrumb trail or url in the results page. Currently lots of the docs I've hit today all were for cloud but theres no way to know until you click into it. Having the breadcrumb/category/url in the result would make it easier at a glance to see what is what so I can zero into content that might help quicker.

Related: #451

AaronSadlerUK:
Added breadcrumbs to the search results with links to the relevant pages

resolves #452

### DIFF
--- a/OurUmbraco/Our/SearchResultExtensions.cs
+++ b/OurUmbraco/Our/SearchResultExtensions.cs
@@ -6,7 +6,6 @@ using System.Web;
 using Examine;
 using OurUmbraco.Our.Models;
 using umbraco;
-using Umbraco.Core.Models;
 using Umbraco.Web;
 
 namespace OurUmbraco.Our
@@ -20,10 +19,30 @@ namespace OurUmbraco.Our
 
         public static List<SearchResultBreadcrumbModel> GetBreadcrumbs(this SearchResult result)
         {
-            var currentResult = UmbracoContext.Current.ContentCache.GetById(result.Id);
-            var ancestors = currentResult.Ancestors().ToList();
-
+            var currentContext = UmbracoContext.Current;
+            var currentResult = currentContext.ContentCache.GetById(result.Id);
             var breadcrumbItems = new List<SearchResultBreadcrumbModel>();
+            if (currentResult == null)
+            {
+                if (result.Fields.Keys.Contains("url") && result["url"].Contains("forum"))
+                {
+                    breadcrumbItems.Add(new SearchResultBreadcrumbModel
+                    {
+                        Name = "Forum",
+                        Url = "/forum/"
+                    });
+
+                    breadcrumbItems.Add(new SearchResultBreadcrumbModel
+                    {
+                        Name = result["nodeName"],
+                        Url = result["url"]
+                    });
+                }
+
+                return breadcrumbItems;
+            }
+
+            var ancestors = currentResult.Ancestors().ToList();
 
             foreach (var ancestor in ancestors)
             {
@@ -58,7 +77,7 @@ namespace OurUmbraco.Our
             {
                 return "solved";
             }
-            
+
             return string.Empty;
         }
 
@@ -114,7 +133,7 @@ namespace OurUmbraco.Our
                     ? string.Format("/{0}/{1}-{2}", url.Trim('/'), result.Fields["__NodeId"], result.Fields["urlName"])
                     : string.Format("/{0}/{1}-{2}.aspx", url.Substring(0, url.LastIndexOf('.')).Trim('/'), result.Fields["__NodeId"], result.Fields["urlName"]);
             }
-            
+
             return "TODO";
         }
 
--- a/OurUmbraco/Our/SearchResultExtensions.cs
+++ b/OurUmbraco/Our/SearchResultExtensions.cs
@@ -1,9 +1,11 @@
 using System;
 using System.Collections.Generic;
 using System.Data.SqlClient;
+using System.IO;
 using System.Linq;
 using System.Web;
 using Examine;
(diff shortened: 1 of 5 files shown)


OurUmbraco/Our/SearchResultExtensions.cs
```diff
@@ -15,6 +18,25 @@ public static string GetTitle(this SearchResult result)
             return HttpContext.Current.Server.HtmlEncode(result.Fields.ContainsKey("nodeName") ? result["nodeName"] : string.Empty);
         }
 
+        public static IEnumerable<SearchResultBreadcrumbModel> GetBreadcrumbs(this SearchResult result)
+        {
+            var currentResult = new UmbracoHelper(UmbracoContext.Current).Content(result.Id) as IPublishedContent;
```
abjerner, line 23: I see a few issues with this line:

- Creating a new `UmbracoHelper` instance has a slight overhead, so initializing more than one per request should generally be avoided. In this line, an instance will be created for each search result. And the search seems to show up to 100 results, meaning a potential of 100 new instances. As our only use of `UmbracoHelper` is getting the IPublishedContent, we can do that directly via `UmbracoContext` instead - eg. by changing the line to:

    ```csharp
    var currentResult = UmbracoContext.Current.ContentCache.GetById(result.Id);
    ```

- Not all search results are based on content. For instance, forum topics are stored in the database, so when attempting to lookup a topic in the content cache, you will get `null` back - which will then result in an exception when trying to get the ancestors of `null` on the next line.

    So ideally, there should be a line line the one below between line 23 and line 24:

    ```csharp
    if (currentResult == null) return new List<SearchResultBreadcrumbModel>();
    ```

    Things don't always go according to plan, so you could also end up with inconsitencies between the content cache and the Examine index - eg. if a content item has been deleted, but still exists in the Examine index. The suggested null check above should also take care of that scenario 😉

OurUmbraco/Our/SearchResultExtensions.cs
```diff
@@ -15,6 +18,25 @@ public static string GetTitle(this SearchResult result)
             return HttpContext.Current.Server.HtmlEncode(result.Fields.ContainsKey("nodeName") ? result["nodeName"] : string.Empty);
         }
 
+        public static IEnumerable<SearchResultBreadcrumbModel> GetBreadcrumbs(this SearchResult result)
```
abjerner, line 21: In the Razor view, you're calling `.ToList()` on the result of this method. Internally in the method it's already a list, so you might as well change the return type the return type to `List<SearchResultBreadcrumbModel>`. and then skip the call to `.ToList()` in the Razor might. That might help make the view look a bit more clean 😉

abjerner (changes requested):
Hi @AaronSadlerUK and thanks for creating this PR 👍 

The search results could certainly do with a breadcrumb, so I had a look at your changes. Generally it seems to work really well, but I left some comments about a few of the lines that could be improved a bit.

As I'm mentioning in one of the comments, forum topics are not stored as content, so they won't be found when you try to do a lookup for them in the content cache. The proposed null check will make sure that users don't see a YSOD or error message, but then forum topics won't have a breadcrumb. So some additional changes should be made for that to happen - but unless you really want to look at tat as well now, I think we can skip that for now, and you or someone else can then have a look at that part later.

AaronSadlerUK:
Thanks for the feedback @abjerner I will look at tweaking this :)

OurUmbraco/Our/SearchResultExtensions.cs
AaronSadlerUK: I have made these changes

OurUmbraco/Our/SearchResultExtensions.cs
AaronSadlerUK: I have made this change

AaronSadlerUK:
I have added the breadcrumbs to the forums result:
image

abjerner:
@AaronSadlerUK Great - I just had a quick look through the code, and it seems to have fixed the issues that I described. I'll see if I can give your PR another test tomorrow ;)

AaronSadlerUK:
I have added the breadcrumbs to the documentation now =]

abjerner (approved):

abjerner:
Great work 👍 

I had did another test, and everything seems to work like described, so I've approved your PR. There isn't really a review program for Our as there is for the CMS, so HQ will have to test and review this as well 😉 

<br /><br /><br /><br />

While testing, I did play a bit with the styling. Personally I think the breadcrumb could do with a slightly smaller text size, and a bit of bottom margin to create some distance to the text below. Eg. turning:

```css
.type-breadcrumb {
    color: #71888e;
    font-size: .95rem;
}
```
into:
```css
.type-breadcrumb {
    margin-bottom: 3px;
    color: #71888e;
    font-size: 0.85rem !important;
}
```

would make:

image

look like below instead:

image

Not really a big thing, so feel free to skip this part if you don't agree 😉

AaronSadlerUK:
> ```css
> margin-bottom: 3px;
> ```

I agree, this is much better!

AaronSadlerUK:
@abjerner I have added modified it to remove the last index from the documentation breadcrumbs.

I'm not sure why this wasn't working before:
```
if (strDirs.Last() == "index")
{
    strDirs.Remove(strDirs.Last());
}
```

nul800sebastiaan:
Lovely work @AaronSadlerUK  and thanks for the thorough review @abjerner - great update! 👍 This will roll out to Our later today.

PR #641, opened 2020-10-16
578 Version filter doesn't appear to be working when searching

AaronSadlerUK:
Fixed the version searching with forum results


AaronSadlerUK:
@nul800sebastiaan is this acceptable for hacktober?

PR #642, opened 2020-10-16
604 If a user isn't on the latest relase page, make it more obvious

### LINKED ISSUE #604: If a user isn't on the latest relase page, make it more obvious
Feature/UX request I think:

When I Google/Bing/DDG "Umbraco v8 release" the page that's returned is [Umbraco V 8.0.0](https://our.umbraco.com/download/releases/800) - which makes total sense, nothing Umbraco can do about it, and the search engine did return me what I asked it to.

It'd be neat if the "also available on Nuget" had some prompt that says "by the way this isn't the latest version", or if the "Latest release" was more prominent on the page. I just unwittingly pasted it straight into the VS Package manager and installed v8.0 into my project instead of 8.6.x 🤦‍♀️.

image

AaronSadlerUK:
Added red text which is shown if not on latest version with a link to the newer version
image

resolves #604

### DIFF
--- a/OurUmbraco.Site/Views/Release.cshtml
+++ b/OurUmbraco.Site/Views/Release.cshtml
@@ -1,4 +1,5 @@
-﻿@using OurUmbraco.Our.Extensions
+﻿@using System.Globalization
+@using OurUmbraco.Our.Extensions
 @using OurUmbraco.Our.Models
 @using OurUmbraco.Our.Services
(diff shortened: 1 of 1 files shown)


AaronSadlerUK:
@nul800sebastiaan is this acceptable for hacktober?

nul800sebastiaan:
Cool, I like it! 👍

Issue #621, opened 2020-09-14
Too long issue title 


joepvtl:
![image](https://user-images.githubusercontent.com/42842226/93056344-95e5fd00-f66c-11ea-913b-7eb6c7c53493.png)

nul800sebastiaan:
Yep! In general it would be lovely if someone could put their designer hat on and help make that list look a little bit better, I've made it functional as best I can now and I would love for it to look a lot better!

umbrabot:
Hi @joepvtl,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly Umbraco GitHub bot :-)

nul800sebastiaan:
Thanks to the wonderful help of @deanleigh I've now updated the release overview to look a whole lot better and deal with wrapping better as well! More updates are needed, but this is (in our humble opinion) quite an improvement already!

deanleigh:
Is this good enough to close the issue?
If it needs more love we could open another?
Not sure what the protocol is.

PR #643, opened 2020-10-17
#621 Release page markup & style changes

lynnewritescode:
- Release page new styling
- Fix alignment issue
- Refactor scss

Before:
before

After:
after

Alignment before:
alignment-before

Alignment after:
alignment-after

### DIFF
--- a/OurUmbraco.Client/src/scss/pages/_release.scss
+++ b/OurUmbraco.Client/src/scss/pages/_release.scss
@@ -1,459 +1,562 @@
-.download.latest .dot {
-	margin-top: 2rem;
-}
+.download {
+  &.latest {
+    .dot {
+      margin-top: 2rem;
+    }
+  }
 
-.download .dot.progress {
-	text-indent: .3rem;
-	text-align: center;
-	line-height: 2.8rem;
-}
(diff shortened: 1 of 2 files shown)


abjerner (approved):
Great work @lynnewritescode - this looks so much better 👍

lynnewritescode:
> Great work @lynnewritescode - this looks so much better 👍

Thank you @abjerner!

nul800sebastiaan:
Thanks @lynnewritescode ! This introduces a few problems that you could not have known about 🙈 The release page has changed in 8.7.0, so the layout is slightly different. I think you can reproduce by creating an 8.7.0 node in the test database (or rename 8.6.0 to 8.7.0). 

With your changes that looks a bit like this:

image


8.6.0 and lower looks great! 

The other thing that this seems to affect is the releases overview:

image

And the compare looks a bit off now as well:

image

I'd love it if you could have a look at this! Note that I did a merge and pushed some commits into your branch, you should be able to do a pull on your branch to get the latest version, let me know if you need help with that.

lynnewritescode:
@nul800sebastiaan Oh no! I can take a look

PR #649, opened 2020-10-30
adding Package.Templates as an umbraco public repository

LottePitcher:
The [Package.Templates repository](https://github.com/umbraco/Package.Templates) now has contribution guidelines, so could it be listed to the list of repos that are considered for home page stats?  If so this PR adds this to the home page links, the json and the GitHub service (i.e. all the places that were updated in this commit: https://github.com/umbraco/OurUmbraco/commit/d9603fd85a274cd332b36daa9bbf2cb35d1e8ec7 when adding umbPack)

Don't know what the order of things should be, but I've added it after UmbPack:

image


nul800sebastiaan:
A little late, but finally merging this now, thanks @LottePitcher - sorry for the long wait!

PR #664, opened 2020-12-13
Fix package listings layout and downloads counts

KevinJump:
Fixes potential layout issue when packages reach 1 million downloads
see : https://github.com/umbraco/Umbraco.Packages/issues/107

this update changes the number format for downloads to display in 1000's (e.g 134.5K) or millions (1.1M) on both the package listing and package info pages. 

image

should also resolve the alignment issue on the package listing page, where the bottom of the boxes doesn't always line up:

For layout, 
image

becomes (the version and download info is now aligned). 

image

### DIFF
--- a/OurUmbraco.Client/src/scss/sections/_packages.scss
+++ b/OurUmbraco.Client/src/scss/sections/_packages.scss
@@ -28,7 +28,7 @@ section.packages {
     }
 
     .package-box {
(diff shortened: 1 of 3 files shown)


jmayntzhusen:
Thanks @KevinJump , works wonderfully 🙂

PR #666, opened 2020-12-28
Set default compatible version to latest

jmayntzhusen:
In combination with https://github.com/umbraco/UmbPack/pull/59 this PR will make it so UmbPack doesn't automatically default to a version the package is compatible with - rather it will default to the latest version of Umbraco Our knows about unless specifically set with the -w param


nul800sebastiaan:
Wow, an old PR! 😅 

With the new marketplace site, we won't be needing this any more. so I'll close this one.

PR #679, opened 2021-02-01
Adding Filip to list of HQ members

bjarnef:
Now that @filipbech is part of Umbraco HQ I think he should be added to the list and add space for new contributors 👍

image


filipbech:
But how will I then earn SWAG going forward 🤣

I'll have to find other ways I guess...

nul800sebastiaan:
Thanks @bjarnef - I just did a grand update of the whole list, including Filip of course, sad to see him disappear from the leaderboard.. 😀

Fake accounts Filip... _hint_

PR #682, opened 2021-03-12
Fix for null reference in nuget download service

dawoe:
Also logging the nuget package id in case this  occurs


nul800sebastiaan:
Thanks @dawoe - will deploy later today!

PR #696, opened 2021-05-04
Add v9 to searchbar dropdown

sofietoft:
As we have started updating the docs for Umbraco 9, we also need a way to search for these new articles.

By adding the v9 option to the searchbar, I hope to have accomplished just that.


sigrundis (changes requested):
Great spot! Almost there ✨ 

Some comments with my thoughts: 

1. Don't you also want to add it to the dropdown here: https://our.umbraco.com/search ? 
Otherwise it just changes to V8 on that page after you use the search feature by clicking "enter" on the keyboard 😊 
You can do so by adding option in this dropdown: 
https://github.com/umbraco/OurUmbraco/blob/main/OurUmbraco.Site/Views/Partials/SearchResults.cshtml#L34

2. It looks like the search controller uses the OurSearcher class for the search which filters by the provided version by using this current major versions list:
https://github.com/umbraco/OurUmbraco/blob/main/OurUmbraco/Our/Examine/OurSearcher.cs#L69 

It seems like you need to add version 9 here for the search to work. 

It also has the functionality to filter by version by excluding other major versions. If we don't include 9 to this list I think we also break that... Because then version 9 won't be excluded when filtered by other versions isn't it?

3. When you fix this I would also fix the variable names, they are very confusing for the version to find for-each statement here:

https://github.com/umbraco/OurUmbraco/blob/main/OurUmbraco/Our/Examine/OurSearcher.cs#L90

I would rename them like so:
```
 //we filter by this version by using the major versions
 var versionsToFind = currentMajorVersions.Where(f => f == versionToFilterBy).ToArray<string>();
 
                foreach (var versionToFind in versionsToFind)
                {
                    sb.AppendFormat("+majorVersion:{0} ", versionToFind);
                }

```
That is, change the variable versionToFind to version**s**ToFind and the variable versionToNegate to versionTo**Find** (looks like a copy-paste mistake from the negate functionality)

4. I think we should change the default current version from taking always the current major version to check first if we have the intex-v<version> parameter in the URL. I think it is weird that the version always changes to v8 after we select the result, even if we search for another version. 

That is, add default value to generate the path parameter index-v<version> and pull the version number from that here: 
https://github.com/umbraco/OurUmbraco/blob/main/OurUmbraco.Site/Views/DocumentationSubpage.cshtml#L7

Happy coding! 👏

sofietoft:
Thanks for the thorough review @sigrundis !

I apologize for not being aware of the extend of adding v9 to the search.
Are you able to make the required changes? I must admit, that I was only able to update the search functionality in that one file, after some clear instructions from my community team 😅 
I'm not a developer, so I'm not entirely sure how I would go about resolving tasks 3-4 on your list here.
I can make PRs for 1 and 2 - it will be in 2 separate PRs though - hope that's OK!

sigrundis:
> Thanks for the thorough review @sigrundis !
> 
> I apologize for not being aware of the extend of adding v9 to the search.
> Are you able to make the required changes? I must admit, that I was only able to update the search functionality in that one file, after some clear instructions from my community team 😅
> I'm not a developer, so I'm not entirely sure how I would go about resolving tasks 3-4 on your list here.
> I can make PRs for 1 and 2 - it will be in 2 separate PRs though - hope that's OK!

Ah I understand, no worries! Great job anyways 😄 However, I think it is best to do everything regarding adding v9 to the search in the same PR - that is, work on it on the same branch so it is possible to test if everything works properly before deploying on dev 😊 

Do you want me to do it on Monday (unfortunately I'm not working today) or do you have another developer who is available to do this?

sofietoft:
Yeah, makes sense to keep it all together 💪 

I'm meeting with the community team today, and they might be able to help me out with what needs to be done here.
If we can't make it today, it would be great if you could look at it on Monday.
I'll shoot you a message before heading off for the weekend!

Thanks! 🙌 
And enjoy your Friday!

sofietoft:
Hi @sigrundis !

After a very productive meeting with docs community team, I think we've managed to resolve everything in this PR!

1 and 2: I've added v9 as a "MajorVersion", and I've included v9 on the general search page as well.
3: After some back and forth, we decided that the naming you've suggested is indeed better than what we had before 😅 So I've updated that!

4: This one, about how the version you've searched for, isn't "stored" or saved", we didn't resolve.
We talked about it, and agreed, that we instead of grabbing the version from the search URL it would make more sense to store it in **HTML Local Storage**. According to the [Privacy Policy on Our](https://our.umbraco.com/privacy-policy) we are already using HTML Local Storage, and thought it would be better (and perhaps even easier 🤔 - to store the version number in there!
How does that sound to you?
This is something that I would have no idea how to do though 😅

Finally, I added an extra thing to this PR.
In this file: https://github.com/umbraco/OurUmbraco/pull/696/commits/64436a03dfdceb55297210801bb554f0975768eb#diff-c0454b430e5d337e25221faaeb40f21e281528bdde999be09db5993a7700db4a

We amended the code for the versioning box (see image below) just a bit.
Before it would ONLY show the box if the article exists in more than 1 version.
The changes we've made will ensure that the box will now always show, whenever a version has been set in the YAML tag, versionFrom.
When we first implemented this versioning system, we assumed that there would always be more than one version of an article. That's not the case, however, and in those cases, it is almost impossible to figure out which version to the article is then targeting 😅 
We hope that by making sure the versioning box will always show, and therefore always tells you which version the article you're reading is targeting, that we can avoid the confusion!

I hope it all makes sense! 😄 
Happy to explain further if needed - just let me know 💪

sigrundis (commented):
Hi! Great job, you can't say that you are not a developer anymore 😉 

Looks like everything regarding the search works now except the dropdown goes back to V8 after navigating to the search result. Should I create a task for it in my backlog or will someone else fix it? 

Regarding always showing the altbox:
1. It is also displayed on the main page, is that intended? If so we at least need to fix the styling, it is blocking the text in the information box.
image
2. I also think the spacing is a bit too narrow, it's a bit crowded... 😬 
image

sofietoft:
Great to hear that the search functionality is now in place.
So, about how to version dropdown switches back - see my comments regarding adding these to the sessions, or using HTML Local storage. I think that would be the best solution to that.
Would be great if you could add it to your backlog, as we will not be able to add the functionality.

About that versioning box.
The reason we started making changes to it, was because we came across come articles that only exist for an older version, but doesn't show the versions box. This makes it look like the article is actually targeting the current version of Umbraco, which is not the case.
So we thought that showing the versioning box whenever the "versionFrom" metadata tag is present on an article, was the way to go.
Looks like we didn't get it quite right though. I'll take a quick look at it with the team today. Do you have an idea of why it's still showing on pages where the version hasn't been set? 🤔 

In any case, this is not something urgent, and it also doesn't make sense that we added it to this PR.
I'll take the changes out of this PR, and create a new one with the changes instead. We'll take a look at it in the team when we have a meeting this afternoon.

sigrundis (approved):
Great, I agree it makes sense to put it in another PR so it is not blocking this task 😊.

Yes I saw your thoughts regarding how to implement the fix for making sure the dropdown is in line with the search, I will put it in my backlog. I will not have time for it in this sprint though but at least it will not be forgotten. 😊

OurUmbraco.Site/Views/Partials/Documentation/DisplayMarkdown.cshtml
```diff
@@ -118,16 +118,12 @@
...
-            else if (currentVersion != null)
-            {
-                <dd class="current-version selected-version">
-                    @(currentVersion.Version)
-                </dd>
-            }
-        </dl>
-    </div>
-</aside>
+
```
nul800sebastiaan, line 121:
```suggestion
            else if (currentVersion != null)
            {
                <dd class="current-version selected-version">
                    @(currentVersion.Version)
                </dd>
            }
```

OurUmbraco.Site/Views/Partials/Documentation/DisplayMarkdown.cshtml
```diff
@@ -118,16 +118,17 @@
                         }
                     </dd>
                 }
-            }
             else if (currentVersion != null)
```
nul800sebastiaan, line 121:
```suggestion
            }
            else if (currentVersion != null)
```

PR #715, opened 2021-07-29
Dev/packages v9 dropdown

KevinJump:
Adds v9 to the drop down on the packages list pages. 

changing the url to  v9 does currently list all the v9 packages, the option is simply missing from the drop down 

see : https://our.umbraco.com/packages/?orderBy=popularity&version=9


AndyButland:
Tested and works as expected.

PR #718, opened 2021-08-04
Remove icon imp text

erikjanwestendorp:
image

unintended text deleted


nul800sebastiaan:
Duh.. 🤦  I saw that before and missed it when I saw the nice big image appear above it! 😅 Thanks again for the help! 🎉

PR #730, opened 2021-10-04
Fixed forum comments from shifting down when hovered over

OwainJ:
The labels on comment action buttons had a `line-height` of 1.3, whilst their icons had a `line-height` of 1, this was causing everything below them to shift slightly down when they became visible.

**Current behaviour:**
OurUmbraco - comment action button line height - old

I fixed this by changing the `line-height` of the labels to 1, so that they match the `line-height` of their icons, and no longer shifts the content down when they become visible.

**Fixed behaviour:**
OurUmbraco - comment action button line height - new

I've tested on Edge, Chrome and Firefox.

### DIFF
--- a/OurUmbraco.Client/src/scss/elements/_forum-thread.scss
+++ b/OurUmbraco.Client/src/scss/elements/_forum-thread.scss
@@ -342,6 +342,7 @@
 
 				span {
 					display: none;
+					line-height: 1;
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Thanks a bunch @OwainJ 👍 and sorry for the wait!

PR #743, opened 2022-02-01
Adds back-office management, display and API updates to support promoted packages in the back-office and on our.umbraco.com.

AndyButland:
This PR adds the option for flagging certain packages as "promoted" (planned for HQ, as a technical partner benefit and to support selected community packages).  By selecting a package as promoted it'll appear in a new section under "Popular" and above "Latest", and also expose the information via the API so we can do a similar thing in the back-office.

Requires a new property on the "Project" document type:

```
Tab: Project
Alias: isPromoted
Name: Is Promoted?
Description: The package is selected for promotion (HQ, technical partner or selected community package).
```

I've reworked the templates slightly to allow for hiding of the whole "Promoted" section if there are none to promote, which required a new partial, and then to share some common helpers and functions between the new and existing template for listing packages.

### DIFF
--- a/OurUmbraco.Site/App_Code/RenderProjectHelpers.cshtml
+++ b/OurUmbraco.Site/App_Code/RenderProjectHelpers.cshtml
@@ -1,0 +1,60 @@
+﻿@using OurUmbraco.Our
+@using OurUmbraco.Repository.Models
+@using OurUmbraco.Project
+
+@helper RenderProjectBox(Package projectContent)
+{
+    <a class="package-box" href="@projectContent.Url">
+        <div class="package-image">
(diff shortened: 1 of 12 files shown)


nul800sebastiaan (changes requested):
Thanks @AndyButland! Before I have a deeper look, I think we need a migration to add the toggle to the doctype, you can just copy this method and update it for your purposes here - https://github.com/umbraco/OurUmbraco/blob/main/OurUmbraco/Our/MigrationsHandler.cs#L85 (yes, it's very poor-man's migrations 😅 ).

AndyButland:
Thanks, have updated with the migration.

PR #753, opened 2022-02-09
Fix #752 - get paged nuget entries when package has 128+ releases.

### LINKED ISSUE #752: Package Average downloads fails when package has more than 128 entries in nuget.
When a package has more than 128 entries in nuget the nuget api returns a different response when you are listing the packages, 

this is mentioned in the current comments for the nuget download service - and defined here :  
 https://docs.microsoft.com/en-us/nuget/api/registration-base-url-resource

as a result when a package has more than 128 entries the code doesn't find any entries and gives it an average downloads per day of zero.

KevinJump:
Fixes the average downloads per day issue on our .


nul800sebastiaan:
Thanks @KevinJump and sorry for the delay, didn't see this until recently! 😅  

I did a bit of cleanup of the code (personal preference) and crucially added a null check as the code stopped working there 14 days ago https://github.com/umbraco/OurUmbraco/pull/753/files#diff-fd08b979a67edfcace0a9365fea9b6a9ae6d42f758101ad9bc03e7106996d0ecR197

Also added some logging to see what's going on and if we have any specific packages that fail in the process. I'll deploy this now and see if it works again 🤞

nul800sebastiaan:
Added a few more fixes, crucially this one https://github.com/umbraco/OurUmbraco/commit/734d516aa9f288eac44a184eb0cfdb26bf400704

Which broke with a NuGet URL of `http://nuget.org/madeup` - understandably. 😅 

That is now fixed and your new code seems to at least not fail.. not sure what I could check, was there any packages with more than 128 releases?

KevinJump:
uSync is the only one i am aware of 😄  - it does appear to be back on the list.

KevinJump:
:) - I have referred to this as the "make uSync great again PR"  in the past 😄  - pre this once you got to 128 downloads your NuGet score was zero, so uSync was then about 10th or 11th on the list .

nul800sebastiaan:
1.81 million is a ridiculous(ly GREAT) amount, well done! 🤩 🤩 🤩

PR #757, opened 2022-03-08
DRAFT work in progress on markdown tabs

marcemarc:
There is much more to do... but this shows the issue with whitespace being removed from code samples when the are in tabs


sofietoft:
Hi 👋 
I'm trying to test this on my local Our install.
When I try to access the Documentation section, I run into this exception/error:
`[FileNotFoundException: Could not load file or assembly 'Markdig, Version=0.16.0.0, Culture=neutral, PublicKeyToken=null' or one of its dependencies. The system cannot find the file specified.]`

I also cannot find the Markdig in the `/bin` folder - which is where I'm assuming it should be?
In VS in `OurUmbraco.Site\Views\Partials\Documentation\DisplayMarkdown.cshtml` line 57 it will me that `class OurUmbraco.Documentation.Businesslogic.MarkdownLogic` is deprecated and obsolete.
In line 58 it's using the path to that class, and it looks to me like this is where it fails 🤔 

Can you see what I'm missing here, in order to get this running?

nul800sebastiaan:
This one became a little stale 😅 
With the new docs site, we won't be needing this any more. so I'll close this one.

PR #758, opened 2022-03-11
product name corrections

mikeirvingweb:
Corrections to product naming, including:

.Net -> .NET, MacOS -> macOS, , , Mac OS -> macOS, SQL server -> SQL Server, Sql Server -> SQL Server, visual studio -> Visual Studio, XCode -> Xcode


nul800sebastiaan:
Thanks.. but none of these files should be updated as they're unused or coming from different sources.

PR #771, opened 2022-05-31
Adding rewrites for moved folders

sofietoft:
In order to not have a folder called "V9-Config" in the documentation, I've had to move the articles around and add a couple of new folders.

This will require a set of rewrite rules, which I've added here.
The rules have been tested on a local instance of Our.

**This shouldn't be merged until the related PR (https://github.com/umbraco/UmbracoDocs/pull/4020) has been merged on the Docs**


sofietoft:
Ran another test with this, and it seems that both menu sections "open" when selecting "Configuration for Umbraco 7 and 8"
image

I suspect this is related to the naming - that both sections are named using "Configuration".

Can someone see a way around this?

PR #797, opened 2022-10-16
fix incorrect wording

LottePitcher:
BTW I know this isn't a Hacktoberfest repo. Am just sending this as I noticed it. Definitely not a Hacktoberfest worthy PR 😆


nul800sebastiaan:
Good spot, thanks @LottePitcher!

PR #810, opened 2022-12-08
Update the meetups groups list

LottePitcher:
I assume the reason why the new 'umbraco-virtual' meetup isn't appearing on the Our home page is because it is missing from this file.

Thought it would be community minded to see who else was missing, and have a quick review of the ones already there:

- Umbraco DK meetup appears to have changed its name, so fixed that
- Denver and RheinMein were in there twice, so removed the duplicates
- meet-umbraco-the-umbraco-meetup (?!) no longer exists so removed

All the following are groups in the Umbraco Pro account but weren't on the list, so I've added them:

- Kochi-umbraco-meetup-group
- North-carolina-umbraco-meetup
- umbracalong
- umbraco-ahmedabad
- umbraco-ireland-meetup
- umbracymru
- wessex-umbraco
- umbraco-virtual


nul800sebastiaan:
🎉 Thank you @LottePitcher !! 💪

PR #817, opened 2023-01-20
Add .NET 6 & 7 to packge .NET version options

callumbwhyte:
_Yes yes, I know we're supposed to be using the awesome new Marketplace now..._

I came back into Our for the first time in ages to update an old package. I noticed that it's possible to mark your package as supporting Umbraco 9, 10, 11, 12... but only possible to mark your package as supporting .NET 5, rather than .NET 6 or .NET 7 as required for Umbraco 10/11 respectively.

### DIFF
--- a/OurUmbraco.Site/Views/Partials/Project/EditFiles.cshtml
+++ b/OurUmbraco.Site/Views/Partials/Project/EditFiles.cshtml
@@ -128,6 +128,8 @@
                                 <option value="4.7.1">4.7.1</option>
                                 <option value="4.7.2">4.7.2</option>
                                 <option value="5.0.0">5.0.0</option>
+                                <option value="6.0.0">6.0.0</option>
+                                <option value="7.0.0">7.0.0</option>
                             </select><br>
                         </p>
 


nul800sebastiaan:
That checkbox list is auto-populated to run to v12, but shouldn't be :-) 

I think we should cut that list off after v8 and leave messages saying packages for v9+ can be created in the new marketplace. There's no zip install of v9+ packages anyway, so it was just a way to promote your nugets. Now that marketplace is in the backoffice's packages section there's no need to keep supporting any packages for v9+.

callumbwhyte:
I thought of this after... I agree. Happy to make a PR if you'll accept one!

nul800sebastiaan:
Thanks! Would be great! Otherwise I'll have a look next week. I remember that version list probably has some horrible code in it, so consider yourself warned! 😅

callumbwhyte:
I finally found some time to take a look at this and the change wasn't too bad...

Whilst there's a number of ways the version list is fetched in codebase, they all end up calling the `UVersion` class one way or another. This already had a restriction for `version > 6` so felt reasonable to add another check for anything below V9 here.

This change does not impact the docs site, the releases page, or the "Compare Releases" feature.

nul800sebastiaan:
Thanks @callumbwhyte - looks like it's all good, I just renamed the file in the csproj as well otherwise it wouldn't build. 😉

callumbwhyte:
Doh! Thanks!

PR #819, opened 2023-03-01
Redirects!

sofietoft:
Added redirects that sends all Cloud and Heartcore docs requests to the new platform.


abjerner:
If it helps anything, I think Sebastiaan installed Skybrud Redirects few months back, which might be easier to handle new redirects. If not, just ignore me 😄

sofietoft:
Hi @abjerner !
We did indeed install that, and it's been working wonders!
But for some reason, it doesn't work for all Cloud or Heartcore paths 🤷‍♀️

PR #823, opened 2023-04-26
Enabled Twitter handle changes

richarth:
Added a Twitter alias field to the profile edit form so members can change their Twitter handle

### DIFF
--- a/OurUmbraco.Site/Views/Partials/Members/Profile.cshtml
+++ b/OurUmbraco.Site/Views/Partials/Members/Profile.cshtml
@@ -134,6 +134,11 @@
                 @Html.TextBoxFor(m => m.Email)
             </div>
 
+            <div class="profile-input" id="twitter">
+                @Html.LabelFor(m => m.TwitterAlias)
+                @Html.TextBoxFor(m => m.TwitterAlias)
+            </div>
+
             <div class="profile-input" id="password">
                 @Html.LabelFor(m => m.Password)
                 @Html.PasswordFor(m => m.Password, new { @class = "password-input" })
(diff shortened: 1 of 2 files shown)


abjerner:
Just for reference, the code has some uncommented logic for linking your Our account with your Twitter account. I created this through a PR a handful of years ago, but can't remember why it's now disabled. But it was disabled in this commit: https://github.com/umbraco/OurUmbraco/commit/04a6205d9de7824124987252f0100a4d1ea13f92

Like with GitHub, the idea was that you could only add an account that you are have access to, so you can't specify a Twitter handle that you don't own. Before this feature was added, there also wasn't any validation to the Twitter handle, so some would just write the alias, some with a leading `@`, and some the full URL. So if the manual field is re-added, this is probably something that should be addressed somehow.

I saw on Discord that Umbraco's access to the Twitter API was revoked, but reading Twitter's docs, scenarios like this (login with Twitter) should still be possible, so it might also be worth re-enabling the **Link with Twitter** functionality if Umbraco can regain their access to the API.

richarth:
Thanks for the feedback.  I did spot your code for doing this properly, but with all the shenanigans going on at Twitter and with some of us needing to add/change our Twitter handles I thought I'd submit a quick fix.

I've updated my PR to strip out any `@` and Twitter URLs before saving the handle.

If HQ are able to restore their Twitter API access I'd be happy to submit a PR to uncomment the linking code.

nul800sebastiaan:
We won't be paying for API access. I'll look at this PR soon though! 👍 Thank  you @richarth!

nul800sebastiaan:
Sorry for the delay @richarth - all good now! I will roll this out soon!

PR #842, opened 2023-11-13
Updated out of date docs links

richarth:
Updated the links to the installation and upgrading docs so they now point at the new docs site


nul800sebastiaan:
Thanks @richarth ! 👍

PR #853, opened 2023-12-15
Remove version number from footer

erikjanwestendorp:
Maybe I am being overly critical, but I think it's a good idea to remove the version number from the footer, especially since Umbraco 7 is end of life.


nul800sebastiaan:
It doesn't really bother me, but indeed it used to be a proud statement to show that it was up to date. Right now, it's a little behind for sure 😅 

Thanks for the PR!

PR #854, opened 2024-01-23
Remove Adolfi.dev from CommunityBlogs

Adolfi:
I am in the process of rebuilding/moving my blog and I have not had time to create an RSS feed yet. And I dont know when I will have the time so until then I will remove myself from this feed.

Will add myself again once I have added a rss. :)


nul800sebastiaan:
Thanks @Adolfi ! 👍

PR #866, opened 2024-12-01
Add Sam Forrest blog (samforrest.dev/blog) to community blogs

SamuelJForrest:
Added my blog (samforrest.dev/blog) to the community blogs.


nul800sebastiaan:
Hey @SamuelJForrest - not sure what's going on but the feed URL is returning a 404 not found, even though it is actually returning the feed content. But at this point the feed reader code has already stopped because it's not returning a 200. 

image


Are you generating this feed from Umbraco? If so, I can dig up some code if you need it!

PR #873, opened 2026-02-10
Change URL and RSS feed for Søren Kottal

skttl:
Updated the URL and RSS feed for Søren Kottal.


mattbrailsford:
Oooh, nice new blog ✨

skttl:
Try doing the konami code :)

mattbrailsford:
Ooooh, psychedelic 🌈

nul800sebastiaan:
Thanks @skttl - sorry for the delay.. we have been moving where the RSS feeds are handled to a different system and you're in that list now. This new list is being picked up by the community site (see below), UMB.FYI and posts to the Discord `#news` channel as well.


image

PR #881, opened 2026-04-21
Update blog details for Søren Kottal

skttl:


skttl:
Oh, sorry - already did it in #873 :)

